PR #3, opened 2019-10-31
Added marketplace and changed files

JorgeVidal29:
Made the market place and changed the login from the index js to its own file

### DIFF
--- a/fortuna/src/Login.js
+++ b/fortuna/src/Login.js
@@ -1,16 +1,15 @@
+//@flow strict
 import * as React from 'react';
 import 'bootstrap/dist/css/bootstrap.min.css';
 import ReactDOM from 'react-dom';
-import MarketPlaceDisplay from './MarketPlace.js';
-const styleForRow = {
-    'margin-top':'20px',
-  };
+import MarketPlace from './MarketPlace.js';
+import './Login.css';
   
   //This is the Login Page Display
   class LoginPageDisplay extends React.Component {
     handleLoginClick = () => {
         LoginFunction();
-      }
+      };
     render(){
       return (
         <div id = "Parent">
@@ -19,12 +18,12 @@ const styleForRow = {
                 <h1 class = "col-md-4 text-center">Fortuna</h1>
                 <h4 class = "col-md-4 text-right">What is Fortuna?</h4>
             </div>
-            <div class = "row" style = {styleForRow}>
+            <div class = "row styleForRow">
               <div class="col text-center">
                   <button type="button" class="btn btn-success btn-lg" onClick={this.handleLoginClick}>Login</button>
               </div>
             </div>
-            <div class = "row" style = {styleForRow}>
+            <div class = "row styleForRow">
               <div class="col text-center">
                   <button type="button" class="btn btn-secondary btn-sm">Signup</button>
               </div>
@@ -40,7 +39,7 @@ const styleForRow = {
   {
     const rootComponent=document.getElementById('root');
     if (rootComponent != null) {
-        ReactDOM.render(<MarketPlaceDisplay/>, rootComponent);
+        ReactDOM.render(<MarketPlace/>, rootComponent);
     }
   }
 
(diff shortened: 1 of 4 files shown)


fortuna/src/Login.js
```diff
@@ -0,0 +1,50 @@
+import * as React from 'react';
```
SecondThread, line 2: We need //@flow strict please :)

fortuna/src/MarketPlace.js
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react';
```
SecondThread, line 2: Definitely //@flow strict here too please

fortuna/src/Login.js
```diff
@@ -0,0 +1,50 @@
...
+import 'bootstrap/dist/css/bootstrap.min.css';
+import ReactDOM from 'react-dom';
+import MarketPlaceDisplay from './MarketPlace.js';
+const styleForRow = {
+    'margin-top':'20px',
+  };
+  
+  //This is the Login Page Display
+  class LoginPageDisplay extends React.Component {
+    handleLoginClick = () => {
```
SecondThread, line 10: This looks like it's a global variable being declared by accident in a method. Use const instead to avoid shooting yourself in the foot later by accident :)

fortuna/src/Login.js
```diff
@@ -0,0 +1,50 @@
+import * as React from 'react';
+import 'bootstrap/dist/css/bootstrap.min.css';
+import ReactDOM from 'react-dom';
+import MarketPlaceDisplay from './MarketPlace.js';
+const styleForRow = {
+    'margin-top':'20px',
```
SecondThread: I'm pretty sure you want ``marginTop: '20px'``

fortuna/src/Login.js
```diff
@@ -0,0 +1,50 @@
...
+  };
+  
+  //This is the Login Page Display
```
SecondThread, line 8: Glad we have that comment to clear things up :)

fortuna/src/MarketPlace.js
```diff
@@ -0,0 +1,31 @@
...
+                    <h2 class = " text-right">Search by Creator</h2>
+                </div>
+            </div>
+          </div>
+      );
+    }
+  
+  }
+
+  export default MarketPlaceDisplay
```
SecondThread: Semicolon (and newline plz)

fortuna/src/MarketPlace.js
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react';
+import 'bootstrap/dist/css/bootstrap.min.css';
+  
+  //This is the MarketPlace Display
+  class MarketPlaceDisplay extends React.Component {
```
SecondThread: The class name needs to match the file name so we don't get super confused when we are importing stuff.

fortuna/src/Login.js
```diff
@@ -0,0 +1,50 @@
...
+  class LoginPageDisplay extends React.Component {
+    handleLoginClick = () => {
+        LoginFunction();
+      }
```
SecondThread: missing semicolon here

SecondThread (changes requested):

fortuna/src/Login.js
SecondThread: Put it in a css file

fortuna/src/Login.js
JorgeVidal29: Const brings an error that it can't be done in javascript

fortuna/src/Login.js
```diff
@@ -0,0 +1,49 @@
...
+              </div>
+            </div>
+          </div>
+      );
+    }
+  
+  }
+
+  //This is the function that gets used when the player clicks login on the dashboard
+  function LoginFunction()
```
SecondThread, line 38: This definitely needs to be fixed long term and is asking for stuff to break. We either need to use components' state/props to do this, or we need a flux store. I'm usually not a fan of merging stuff into master that is really bad practice, but I guess this stuff will probably be fixed when the login section gets handled, so it would make things easier probably.

SecondThread (approved):

PR #4, opened 2019-11-05
Adam1

blaiadam:
Merged Jorge and my branches together. Also connected the webpages.


fortuna/src/App.js
```diff
@@ -0,0 +1,17 @@
+//@flow strict
+
+import * as React from 'react';
```
SecondThread: I like the abstraction :)

fortuna/src/login/Login.js
```diff
@@ -0,0 +1,54 @@
+//@flow strict
+
+import * as React from 'react';
+import ReactDOM from 'react-dom';
+import 'bootstrap/dist/css/bootstrap.min.css';
+import LoginPopup from './LoginPopup.js';
+import MainMenu from '../mainmenu/MainMenu.js';
+
+const styleForRow = {
+  'margin-top':'20px',
```
SecondThread: I'm pretty sure we should use `marginTop: '20px'` instead. Building variable names as strings is messy and I think we should avoid it. See the accepted answer here for an example of different things you can do (but I strongly prefer marginTop personally):
https://stackoverflow.com/questions/49825386/how-to-add-style-like-margin-to-react-component

fortuna/src/login/LoginPopup.js
```diff
@@ -0,0 +1,46 @@
...
+import 'bootstrap/dist/css/bootstrap.min.css';
+import Popup from 'reactjs-popup';
+
+// Informational popup for login page.
+class LoginPopup extends React.Component<{||}> {
+  render(): React.Node {
+    return(
+	    <Popup trigger={<button className="button">What is Fortuna?</button>} modal>
+	      {close => (
+	        <div className="modal">
```
SecondThread: Can we fix the formatting and indentation here please?

SecondThread: Some stuff is really indented, the indentation for the opening and closing tags doesn't match, and there are a bunch of blank lines that confuse me.

fortuna/src/mainmenu/MainMenu.js
```diff
@@ -0,0 +1,72 @@
...
+  };
+
+  handleLoginClick = () => {
+    LoginFunction();
+  };
+  
+  render(): React.Node {
+    return (
+      <div id = "Parent">
+	      <div class = "row">
```
SecondThread: Ditto, clean up indentation please

fortuna/src/mainmenu/MainMenu.js
```diff
@@ -0,0 +1,72 @@
...
+          <div class = "col-md-4"><button type="button" class="btn btn-secondary btn-lg" onClick={this.handleLoginClick}>&lt;- Back to Login</button></div>
```
SecondThread: I believe you could also do '<-' with JSX to avoid the escape codes, but this works

fortuna/src/mainmenu/MainMenu.js
```diff
@@ -0,0 +1,72 @@
...
+            <button type="button" class="btn btn-info">Training Arena</button>
+          </div>
+        </div>
+        <div class="row" style={styleForRow}>
+	        <div class="col text-center">
+            <button type="button" class="btn btn-info" onClick={this.handleMarketplaceClick}>Marketplace</button>
+          </div>
+        </div>
+        <div class="row" style={styleForRow}>
+	        <div class="col text-center">
```
SecondThread: Definitely clean up indentation please :)

fortuna/src/marketplace/Marketplace.js
```diff
@@ -0,0 +1,60 @@
...
+            <h6 class = "text-right">Tank One Jammer</h6>
+            <h6 class = "text-right">Tank One Base</h6>
+          </div>
+        </div>
+      </div>
+    );
+  }
+}
+
+function MainMenuFunction() {
```
SecondThread: This REALLY sketches me out. React uses a top-down approach. If we need to switch what is rendered, it needs to be passed in as a prop. Editing the DOM tree is definitely not something we should be doing in any component other than the root index.js component. If we have multiple components simultaneously (or put the main menu in a wrapper class, for example), this will just delete all of them. Instead we need to store which tab we are looking at in the App component's state, and then change that by either calling a flux store or calling a function that was passed in as a prop from app.js or something.

fortuna/src/mainmenu/MainMenu.js
```diff
@@ -0,0 +1,72 @@
...
+            <button type="button" class="btn btn-info">Armory</button>
+          </div>
+        </div>
+      </div>
+	  );
+  }
+}
+
+// Handles movement to the Marketplace.
+function MarketplaceFunction() {
```
SecondThread: I really disagree with this way of doing things. I explain why at the bottom of the code review.

fortuna/src/mainmenu/MainMenu.js
```diff
@@ -0,0 +1,72 @@
...
+  const rootComponent = document.getElementById('root');
+  if (rootComponent != null) {
+    ReactDOM.render(<Marketplace/>, rootComponent);
+  }
+}
+
+// Handles movement to the Login page.
+function LoginFunction() {
```
SecondThread: ditto

fortuna/src/login/Login.js
```diff
@@ -0,0 +1,54 @@
...
+          <div class = "col text-center">
+            <LoginPopup/>
+          </div>
+        </div>
+	    </div>
+	  );
+  }
+}
+
+function MainMenuFunction() {
```
SecondThread: not a fan, explained at bottom

fortuna/src/App.js
```diff
@@ -0,0 +1,17 @@
...
+import 'bootstrap/dist/css/bootstrap.min.css';
+import Login from './login/Login.js';
+
+// This builds the page.
+class App extends React.Component<{||}> {
+
```
SecondThread: This needs to store what we are rendering somewhere in its state, and then pass to children a callback method for state change at least if we aren't using flux stores.

SecondThread (changes requested):

fortuna/src/login/Login.js
SecondThread: @blaiadam Can you address this?

fortuna/src/login/Login.js
```diff
@@ -0,0 +1,44 @@
...
+              <button type="button" class="btn btn-success btn-lg">Login</button>
+            </Link>
+      	  </div>
+      	</div>
+      	<div class = "row" style = {styleForRow}>
+      	  <div class="col text-center">
+      	    <button type="button" class="btn btn-secondary btn-sm">Signup</button>
+      	  </div>
+        </div>
+        <div class = "row" style = {styleForRow}>
```
SecondThread: Can you clean up the indentation here before pushing too please :) Thanks

fortuna/src/marketplace/Marketplace.js
```diff
@@ -0,0 +1,51 @@
...
+  
+// Marketplace component.
+class Marketplace extends React.Component<{||}> {
+	
+  render(){
+    return (
+      <div id = "Parent">
+        <div class = "row">
+          <div class = "col-md-4">
+		    <Link to="/MainMenu">
```
SecondThread: That's super clean and looks amazing. Thanks for fixing it!

SecondThread (approved):
Looks much better. Please address the commends I made in the earlier review before pushing though :)

fortuna/src/backend/routes/api/signup.js
```diff
@@ -0,0 +1,38 @@
+//@flow strict
```
SecondThread, line 1: Did you make the changes to this file? If not, they shouldn't be in this PR. If someone else made this in some other commit, they should push those themself in a separate PR. If we conglomerate all the changes into one branch and push that, then we lose track of who wrote which code, so we can't keep track of who knows how what works, et cetera

fortuna/src/casus/BlockBank.js
```diff
@@ -0,0 +1,30 @@
+//@flow strict
```
SecondThread, line 1: These changes definitely shouldn't be part of this PR.

fortuna/src/casus/Casus.js
```diff
@@ -0,0 +1,32 @@
+//@flow strict
+
```
SecondThread, line 2: Same as earlier, this stuff should be merged individually

fortuna/src/casus/blocks/BoundingBox.js
```diff
@@ -0,0 +1,19 @@
+//@flow strict
+
```
SecondThread, line 2: Okay, this stuff is going to cause all sorts of merge conflicts when you merge it I think. You might have to undo some of the commits merging stuff in that you didn't write so we don't have two copies everywhere

SecondThread (changes requested):

PR #6, opened 2019-11-11
Added api route to call user signup

Airor4:


fortuna/src/backend/routes/api/home.js
```diff
@@ -0,0 +1,8 @@
+const express = require('express');
```
Derplime, line 3: Add //@flow strict at the top for when we use flow to check files.

See the server.js for an example

fortuna/src/backend/routes/api/signup.js
```diff
@@ -0,0 +1,39 @@
+const express = require('express');
```
Derplime, line 3: //@flow strict at the top here too

Derplime (commented):
Just a couple of things to add before my approval.

Derplime:
Did the flow changes locally and ran flow, no errors with the code.

Also tested the api locally, and it works on my machine.

Derplime (approved):
To reiterate, tested API on local machine, and it worked. Also no errors when running flow on it.

PR #10, opened 2019-11-18
Baylor

### LINKED ISSUE #17: Add authentication and login api
Authentication and login API

Airor4:
added JSON Web Token(JWT) authentication and login api route. WAIT NEED TO ADD FLOW.

Closes #17


fortuna/src/backend/routes/api/auth.js
```diff
@@ -0,0 +1,86 @@
+const express = require('express');
```
SecondThread, line 4: Can we get flow here please?

fortuna/src/backend/routes/api/auth.js
```diff
@@ -0,0 +1,86 @@
...
+                .json({ errors: errors.array() });
+        }
+
+    // Assigns request body to user schema fields
+    const { userName, password } = req.body;
+
+    try{
+        // See if User already exists
+        let user = await User.findOne({ userName });
+        if(!user){
```
SecondThread: That's a sketchy null check. `user==null` preferred and it will check for both null and undefined for you

fortuna/src/backend/routes/api/auth.js
```diff
@@ -0,0 +1,86 @@
...
+const jwt = require('jsonwebtoken');
+const { check, validationResult } = require('express-validator');
+const config = require('config');
+const bcrypt = require('bcrypt');
+
+
+const User = require('../../../models/userModel');
+
+
+router.get('/', auth, async (req, res) => {
```
SecondThread: What types are req and res expected to be? specifying this with flow would be helpful I think

fortuna/src/backend/routes/api/auth.js
```diff
@@ -0,0 +1,86 @@
...
+
+// Authenticate user & get token
+// Public
+router.post('/', [
+    check('userName', 'Please enter a username with 3 or more characters')
+        .isLength(3),
+    check('password', 'Password is required')
+        .exists()
+    ], 
+    async (req, res) => {
```
SecondThread: Ditto: can we get types on req and res so stupid people like me know what's going on?

fortuna/src/backend/routes/api/auth.js
```diff
@@ -0,0 +1,86 @@
...
+                .status(400)
+                .json({ errors: errors.array() });
+        }
+
+    // Assigns request body to user schema fields
+    const { userName, password } = req.body;
+
+    try{
+        // See if User already exists
+        let user = await User.findOne({ userName });
```
SecondThread, line 61: Can this be const or am I just blind here?

fortuna/src/backend/routes/api/auth.js
```diff
@@ -0,0 +1,86 @@
...
+        const payload = {
+            user: {
+                // Primary key id generated by Mongo
+                id: user.id
+            }
+        }
+
+        // expiresIn 1 hour == 3600 seconds
+        jwt.sign(payload, config.get('jwtSecret'), 
+        { expiresIn: 3600},
```
SecondThread: Are people going to have to re-login every hour? I guess it's a judgement call here, so if you think that's cool, then go for it, but I can imagine people playing for more than an hour.

fortuna/src/backend/routes/api/signup.js
```diff
@@ -3,38 +3,81 @@
...
+                .json({ errors: errors.array() });
+        }
+
+    // Assigns request body to user schema fields
+    const { userName, password } = req.body;
+
+    try{
+        // See if User already exists
+        let user = await User.findOne({ userName });
+        if(user){
```
SecondThread: sketchy non-null check. `user != null` preferred

fortuna/src/backend/routes/api/signup.js
```diff
@@ -3,38 +3,81 @@
...
         }
 
-        
-    });
+        // Need to manage the config path so that the jwttoken is not here see require('config')
+        // expiresIn 1 hour == 3600 seconds
+        jwt.sign(payload, config.get('jwtSecret'), 
+        { expiresIn: 3600},
+        (err, token) => {
+            if(err) throw err;
```
SecondThread: Sketchy null check again

fortuna/src/backend/routes/middleware/auth.js
```diff
@@ -0,0 +1,22 @@
+const jwt = require('jsonwebtoken');
+const config = require('config');
+
+module.exports = function(req, res, next){
```
SecondThread: In general it's cleaner to declare the function earlier (something like `function auth(req: typeOfReq, res: typeOfRes, next: typeOfNext): functionReturnType {`

and then have `export default auth;` here. But either way, it is really helpful for everyone later on if you label the types of what this function takes in and returns at least.

fortuna/src/backend/routes/middleware/auth.js
```diff
@@ -0,0 +1,22 @@
+const jwt = require('jsonwebtoken');
```
SecondThread, line 3: Need `//@flow strict` please :)

SecondThread (changes requested):
If you could get some more flow integration here so everyone else can see how these methods work without reading through all the code, that would be fantastic.

Specifically, label the types of what things are on all the methods that return stuff, and make sure to have `//@flow strict` on all your files so stupid mistakes get caught.

fortuna/src/backend/routes/api/auth.js
Derplime: > What types are req and res expected to be? specifying this with flow would be helpful I think

I would think it means request and response returned from the async anonymous method. But yeah at least clarifying on these would be nice.

fortuna/src/backend/routes/api/auth.js
Airor4: changed to 4 hrs

fortuna/src/backend/routes/api/auth.js
Airor4: A new instance of user is created if one does not exist in the database already.

SecondThread (changes requested):
The node_modules stuff can't be tracked by github. You have 53,000 lines of code that you are adding to the repo here, changes 342 files, and have a scroll bar on the review page that is as small as it can be.

Derplime:
> The node_modules stuff can't be tracked by github. You have 53,000 lines of code that you are adding to the repo here, changes 342 files, and have a scroll bar on the review page that is as small as it can be.

So this is the potential problem that happens with the thing that I suggested about bringing the entire repository up a level. The reason this happened is because @Airor4 added his modules from the `Fortuna/` directory, not the `Fortuna/fortuna/` directory.

PR #12, opened 2019-11-18
New branch merges master and adam1.

blaiadam:


SecondThread (approved):
LGTM, white space on login.js is a little sketchy looking, but we can fix that some other time, it doesn't really matter.

Ship it! :)

PR #30, opened 2019-12-12
Jorge

### LINKED ISSUE #29: Connect Login with Login Api
The login page needs to call the login api and then move on depending on the response. It also needs to have a username and password field.

JorgeVidal29:
Redesign of login page. Has the hooks for the API but want to get with back end team to talk about targeting.

Resolves #29


fortuna/src/Login.js
```diff
@@ -1,49 +1,108 @@
 //@flow strict
 import * as React from 'react';
+import { Link } from 'react-router-dom';
 import 'bootstrap/dist/css/bootstrap.min.css';
-import ReactDOM from 'react-dom';
-import MarketPlace from './MarketPlace.js';
+import LoginPopup from './LoginPopup.js';
 import './Login.css';
-  
+const PORT = 3000;
```
SecondThread: This works for now since this is the only API call, but this should probably get put in some constants file somewhere if it is going to be used all over

fortuna/src/Login.js
```diff
@@ -1,49 +1,108 @@
...
+    };
+    
+    handleLoginClick = async e => {
+      e.preventDefault();
+      const response = await fetch('http://localhost:'+PORT+'/backend/routes/api/signup.js', {
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+        },
+        body: JSON.stringify({ userName: this.state.userName, password:this.state.password }),
```
SecondThread: uh? You sure this passes flow? This seems super sketch since those things might not exist...

fortuna/src/Login.js
```diff
@@ -1,49 +1,108 @@
...
-        LoginFunction();
-      };
-    render(){
+  class Login extends React.Component<{||}> {
+    state = {
+      response: '',
+      userName: '',
+      password: '',
+      responseToPost: '',
+    };
```
SecondThread: The <{||}> part of the `React.Component<{||}>` means that you have no props and no state. But you are using and relying on your state being not empty.

As a side note, it's probably a better idea to set the state in the constructor, as demoed on section 5.2 in the react docs here: https://reactjs.org/docs/state-and-lifecycle.html

The reason this is the case is because you are setting state like it is a class variable. This isn't a thing that should usually exist for components. The class variables could be updated, but the component won't re-render them.

SecondThread: The <{||}> part of the `React.Component<{||}>` means that you have no props and no state. But you are using and relying on your state being not empty.

As a side note, it's probably a better idea to set the state in the constructor, as demoed on section 5.2 in the react docs here: https://reactjs.org/docs/state-and-lifecycle.html

The reason this is the case is because you are setting state like it is a class variable. This isn't a thing that should usually exist for components. The class variables could be updated, but the component won't re-render them.

fortuna/src/Login.js
SecondThread: You sure this passes flow? This seems super sketch since those things might not exist...

fortuna/src/LoginPopup.js
```diff
@@ -0,0 +1,36 @@
+
```
SecondThread, line 1: Flow plz

fortuna/src/Login.js
```diff
@@ -1,49 +1,108 @@
...
+      const response = await fetch('http://localhost:'+PORT+'/backend/routes/api/signup.js');
+      const body = await response.json();
+      if (response.status !== 200) throw Error(body.message);
+      
+      return body;
+    };
+    
+    handleLoginClick = async e => {
+      e.preventDefault();
+      const response = await fetch('http://localhost:'+PORT+'/backend/routes/api/signup.js', {
```
SecondThread: This URL string should definitely be in a helper method somewhere. Reason: we don't want to add a bunch of these and then forget to change one when we switch to the real server

fortuna/src/Login.js
```diff
@@ -1,49 +1,108 @@
...
+      responseToPost: '',
+    };
+    
+    componentDidMount() {
+      this.callApi()
+        .then(res => this.setState({ response: res.express }))
+        .catch(err => console.log(err));
+    };
+    
+    callApi = async () => {
```
SecondThread: This is basically an async function, so you should definitely specify what the return type is (especially since it isn't void).

SecondThread (changes requested):
I pulled this branch locally and got 9 flow errors when I ran `npm run flow`. Most of these came from not giving the types of things and from not typing your state. Pushing this would break everything else's safety checks. There are also files missing //@flow strict

fortuna/src/Login.js
JorgeVidal29: Yup I agree only here to test how we are going to integrate the backend and frontend

fortuna/src/Login.js
JorgeVidal29: You mean username and password may not exist? If thats the case it does in the state as empty by default and I will make note to double check it when handleLoginClick is being called

fortuna/src/Login.js
JorgeVidal29: Ok I deleted the <{||}>. I don't really need these things to be re-rendered as I am never showing them to the user.

fortuna/src/Login.js
JorgeVidal29: Ok. Will do as soon as we figure out if this is how we are doing this

fortuna/src/Login.js
SecondThread: state.userName might be undefined if you only look at the types of things. Understanding that state.userName won't be undefined requires someone reading and understanding all of your code

fortuna/src/Login.js
SecondThread: You always want things to be rerendered; that's the main benefit of the React. Instead of deleteing the types entirely, you have to say what the types are. (namely, don't make your state and props empty, make them contain exactly the information you expect them to have)

PR #31, opened 2019-12-13
Merge Casus Highlighting on Click

SecondThread:


Derplime (approved):
Alright, looks great!

The light grey selection border on a white background was a bit hard to see though. But assuming the background won't be white, it works. Though I would think this PR is more for the functionality and the fine tuning will be done later.

PR #48, opened 2020-01-08
If Block, If-else Block, and While Block for Casus

SecondThread:
**Description**
Added If Block, If-Else block, and While blocks for Casus

**Type of change**

**Checklist:**

- [n/a] I have made corresponding changes to the design document (if applicable)


Derplime (approved):
Looks good to me. Also toyed around with randomly placing blocks where I could as you saw when I mentioned the bit about having something like a scroll if the code goes too far down.

PR #54, opened 2020-01-11
Added Casus Inventory field to user model.

### LINKED ISSUE #52: Add Casus inventory to user model
Upon request, there should be a field for Casus inventory that gives a count of how many types of Casus blocks an individual has.

The original idea of having a DB collection of Casus blocks, with references to the owners of each block was a good idea, but the team pointed out an issue where it's possible to have a lot of users with an insane amount of Casus blocks that they own. In this case, the collection would be absurdly large if we go with the current idea.

**Additional context**
If we move forward with this idea in the PR, we may need to consider the purpose of having the Casus block model file at all. A discussion on how we consider them "existing" in the system may come into play.

Derplime:
**Description**
Adds Casus Inventory field to user model. I went with the inventory being an Object and has the fields that represent the number of blocks for each. If this is different than what you feel should be implemented, I'm open to changing it. 

Main things to check:
- @SecondThread check if I have all your intended blocks listed, I added them based on what made sense from their file names.
- See if the implementation is suitable or if I should change it to another form.

Fixes #52 

**Type of change**

- [x] This change requires a update in the design document (field isn't documented)

**Checklist:**

- [x] I have made corresponding changes to the design document (**Will make them on final approval**)

### DIFF
--- a/fortuna/src/models/userModel.js
+++ b/fortuna/src/models/userModel.js
@@ -34,6 +34,53 @@ const User = new Mongoose.model('User', new Mongoose.Schema ({
     favoriteTankId: {
         type: String
     },
+    // Casus inventory
+    casusInventory: {
+        andBlocks: {
+            type: Number,
+            default: 0
+        },
+        forBlocks: {
+            type: Number,
+            default: 0
+        },
+        ifBlocks: {
+            type: Number,
+            default: 0
+        },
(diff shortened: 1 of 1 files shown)


SecondThread (approved):
Looks good for now. Obviously we will have to add more of these later, but this should be helpful for testing.

PR #57, opened 2020-01-16
CSS Changes

blaiadam:
**Description**
Created a Main.css for us to add and edit to for all web pages. Also cleaned up the file for the Login references by removing class names with the same styling. Please use the colors to update your front end pages.

Fixes #(issue)

**Type of change**

**Checklist:**

- [ ] Any dependent changes have been merged and published in downstream modules


fortuna/src/Main.css
```diff
@@ -0,0 +1,89 @@
+.styleForRow {
+    margin-top:20px;
+}
+
+body, html {
```
SecondThread, line 5: This part is a big change which is why I am looking at it particularly carefully

fortuna/src/Main.css
```diff
@@ -0,0 +1,89 @@
+.styleForRow {
+    margin-top:20px;
+}
+
+body, html {
+    height: 100%;
+    color: #04CCFF;
```
SecondThread, line 6: What does height: 100% do?

Also, what color is this setting?

fortuna/src/Main.css
```diff
@@ -0,0 +1,89 @@
...
+
+
+/* Popup Styling */
+.popup {
+    padding: 5px 50px 20px 50px;
+    background: #04213F;
+}
+
+/* Leaderboard */
+.leaderboard {
```
SecondThread, line 82: IMO this shouldn't be in the CSS file that everyone is using unless we are going to have the leaderboard on every page. Maybe make a seperate leaderboard.css file?

blaiadam: There will be a leaderboard on the Battle Arena page as well. I can make a separate CSS file for it if everyone thinks it's for the best, but I don't think it is necessary.

fortuna/src/Main.css
blaiadam: Sorry, I missed removing the height setting. It was leftover from some testing. I'll fix it now. The color setting changes the color of the text in the container.

fortuna/src/Main.css
SecondThread: I definitely think we should make Leaderboard its own component then, that way we can reuse the logic for it. This is a perfect example of things components are good for.

I don't really care if we make it is own CSS file or not. If we do make Leaderboard its own component though, then it will only be used in one place. I'm totally cool if that's something for another PR though.

SecondThread (approved):

PR #58, opened 2020-01-16
Redesigned Marketplace

JorgeVidal29:
**Description**
Redesigned Marketplace

Fixes #(issue)

**Type of change**

**Checklist:**


fortuna/src/Main.css
```diff
@@ -0,0 +1,106 @@
+.styleForRow {
```
SecondThread: Why are there a bunch of additions on this file if it already existed in master? It makes it hard to see what to focus on when reviewing code.

fortuna/src/Main.css
```diff
@@ -0,0 +1,106 @@
...
+
+.middle {
+    width: 50%;
+    text-align: center;
+    margin-top: 350px;
+}
+
+
+/* Button Styling */
+.loginbtn {
```
SecondThread: If this is in a global CSS file, then these styles should be styles of all buttons. That is, we should have like one main button (like maybe blue or something) and one secondary button (for 'cancel' or whatever). But we definitely shouldn't have stuff specific to login in a global CSS file.

fortuna/src/Main.css
```diff
@@ -0,0 +1,106 @@
...
+    color: #000921;
+    background: #04CCFF;
+    border: 2px solid #04CCFF;
+    border-radius: 8px;
+    width: 25%;
+}
+
+.loginbtn:hover {
```
SecondThread: Same as above

fortuna/src/Main.css
```diff
@@ -0,0 +1,106 @@
...
+	color: #04CCFF;
+	background: #000921;
+	border: 2px solid #04CCFF;
+}
+
+.signupbtn {
```
SecondThread: Same as above; this should be reusable so we don't need new CSS for every button in this entire app

fortuna/src/Main.css
```diff
@@ -0,0 +1,106 @@
...
+    color: #04CCFF;
+    border: none;
+    background: none;
+}
+
+.signupbtn:hover {
```
SecondThread: same as above, but you get the point, so I'll stop

fortuna/src/Main.css
```diff
@@ -0,0 +1,106 @@
...
+    background: none;
+}
+
+.infobtn:hover {
+    color: #39a9dd;
+}
+
+
+/* Popup Styling */
+.loginpopup {
```
SecondThread: All these popups are the same. There is no reason we should have 3 different classes that all mean the same thing and are all used for the same thing.

fortuna/src/Main.css
```diff
@@ -0,0 +1,106 @@
...
+.leaderboard {
+	padding-left: 10px;
+	margin-right: 200px;
+	background-color: #04213F;
+	border: 5px solid black;
+	border-radius: 8px;
+}
+
+.list-group-item list-group-item-action{
+    color:"black";
```
SecondThread: Put a space after your colon please

fortuna/src/marketplace/AccessoriesView.js
```diff
@@ -0,0 +1,13 @@
+//@flow strict
+import * as React from 'react';
+import 'bootstrap/dist/css/bootstrap.min.css';
+
+class AccessoriesView extends React.Component<{||}> {
+    render(){
+		return (
```
SecondThread: Fix the indentation here please

fortuna/src/marketplace/CannonsView.js
```diff
@@ -0,0 +1,13 @@
+//@flow strict
+import * as React from 'react';
+import 'bootstrap/dist/css/bootstrap.min.css';
+
+class CannonsView extends React.Component<{||}> {
+    render(){
+		return (
```
SecondThread: same as above, and same for the rest of the files

fortuna/src/marketplace/AccessoriesView.js
```diff
@@ -0,0 +1,13 @@
...
+            <h2>This is the Accessories View</h2>
+            );
+	  }
```
SecondThread: indentation with these two lines is off too

fortuna/src/marketplace/CasusBlocksView.js
```diff
@@ -0,0 +1,13 @@
+//@flow strict
+import * as React from 'react';
+import 'bootstrap/dist/css/bootstrap.min.css';
+
+class CasusBlockView extends React.Component<{||}> {
```
SecondThread: Just curious: what is the Casus Block View going to show? The Casus stuff is going to be in the CasusContainer component.

fortuna/src/marketplace/Marketplace.js
```diff
@@ -3,9 +3,58 @@
...
+import CasusBlocksView from './CasusBlocksView.js';
+import JammersView from './JammersView.js';
+import TreadsView from './TreadsView.js';
+import AccessoriesView from './AccessoriesView.js';
+import ScannersView from './ScannersView.js';
+
+
+type Props = {||}; 
+type State = {|
+	marketplaceView: React.Element<any>,
```
SecondThread: This is very bad for many reasons.

First: It isn't a good idea to use any, or object unless you absolutely have to. It isn't 'any' type; we know what type it is.

Second (and more importantly): This destroys how react works. Storing the component that we want to render in our state makes it so that when react components get re-rendered the internal components don't actually get re-rendered. The state of the component should be the _information_ about what gets rendered (such as which tab is selected). It shouldn't be the _thing itself_ that is rendered.

Here are some problems that creating components and storing them in your state instead of creating them in the render method causes:
- render() won't act as expected when components are re-rendered. 
- If the props/state of this component update, we will likely need to update the props of a child state. We can't do this if that is stored in our state, without really complex logic in OnComponentDidUpdate.
- We might not know what props we need to pass to this child component until we are in the render method. We are creating the component that we will end up rendering sometime way later and storing it in our state. What do we pass to that component when we create it?

Instead of storing the component, you could store the selectedTankPartType. Then, when the component is actually rendering, you can create the SomethingView that you need to create.

fortuna/src/marketplace/Marketplace.js
```diff
@@ -3,9 +3,58 @@
...
-class Marketplace extends React.Component<{||}> {
+class Marketplace extends React.Component<Props, State> {
+
+	constructor(){
+		super();
+		this.state={
+			marketplaceView: <h2>Select a type on the left</h2>
+		}
+	}
+	showMarketView(selectedTankPartType: string) {
```
SecondThread: If there are a fixed number of strings that this is allowed to be, this needs to be a type (i. e. a type that isn't string). Why is this bad? What if someone called this method and passed in 'Cannons' for example? Flow would say that is just fine, but that clearly isn't acceptable input. The fact that it isn't should be indicated by its method signature.

SecondThread: Also you need to label the return type of this method. (looks like void)

fortuna/src/marketplace/Marketplace.js
```diff
@@ -21,31 +70,25 @@ class Marketplace extends React.Component<{||}> {
...
-						<h6 className="text-right">Tank One Scanner</h6>
-						<h6 className="text-right">Tank One Jammer</h6>
-						<h6 className="text-right">Tank One Base</h6>
+						<div className="text-center">{this.state.marketplaceView}</div>
 		  			</div>
 				</div>
 	  		</div>
 		);
-  	}
+	  }
```
SecondThread: I can't tell if this line used to have bad indentation or now has bad indentation, but it doesn't line up with the diff, so make sure it is right now.

SecondThread (changes requested):

Derplime:
While also doing the changes proposed by @SecondThread, you'll need to update your branch to the new file hierarchy, since that was recently merged in.

fortuna/src/Main.css
JorgeVidal29: Hey for the changes, the file hadn't been put into master yet when I was working. I can delete mine. The login btn is probably a adam question.

fortuna/src/marketplace/CasusBlocksView.js
JorgeVidal29: This is going to be the view for selling stuff that does with Casus on the marketplace that's why its under the marketplace folder

fortuna/src/Main.css
blaiadam: Jorge, we should collaborate on the CSS stuff sometime soon.

fortuna/src/marketplace/ChassisView.js
```diff
@@ -0,0 +1,13 @@
+//@flow strict
+import * as React from 'react';
+import 'bootstrap/dist/css/bootstrap.min.css';
```
blaiadam: So with the way that our web page is functioning, as long as this CSS import is included in Login.js or App.js, it will be imported on all of our js files. This renders the import of the bootstrap CSS file here unnecessary. You should be able to safely remove this from all of the files in this PR.

SecondThread (changes requested):
You gotta pull from master bud. We don't want 108 files changed from files you haven't touched.

image

run
git pull origin master

then push again

PR #119, opened 2020-03-02
Updated/Added Models to reflect CDR

### LINKED ISSUE #112: DB Model Refresh/Update 1: The Model Menace
There were some models that were updated as well as added to reflect the ERD we presented in CDR. Just need to add and update those.

N/A

**Additional context**
N/A

Derplime:
**Description**
Updated some models and added some to reflect what we discussed for CDR.

Resolves #112 

**Type of change**

- [x] This change requires a update in the design document

Passes Flow

**Final Checklist:**

### DIFF
--- a/src/models/battleRecordModel.js
+++ b/src/models/battleRecordModel.js
@@ -1,0 +1,73 @@
+//@flow strict
+// Document Schema for the Battle Record
+// Used for Battles and Replays
+
+// Required Imports
+import Mongoose from 'mongoose';
+
+const BattleRecord = new Mongoose.model('BattleRecord', new Mongoose.Schema({
+    // _id field made upon creation
+    
+    // ID of users
+    userOne: {
+        type: String,
+        required: true,
+        ref: 'user'
+    },
+    userTwo: {
+        type: String,
+        required: true,
+        ref: 'user'
+    },
+    // 0 = Tie, 1 = User one, 2 = User two
+    winner: {
+        type: Number,
+        required: true,
+        enum: [0, 1, 2]
+    },
+    // Total of money wagered
+    prizeMoney: {
+        type: Number,
+        required: true
+    },
+    // User one's tank
+    tankOne: {
+        tankName: {
+            type: String,
+            required: true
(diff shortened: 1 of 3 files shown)


src/models/battleRecordModel.js
```diff
@@ -0,0 +1,73 @@
...
+        type: String,
+        required: true,
+        ref: 'user'
+    },
+    userTwo: {
+        type: String,
+        required: true,
+        ref: 'user'
+    },
+    // 0 = Tie, 1 = User one, 2 = User two
```
SecondThread: Just thought of this: is the front end going to tell you what the tanks of the two players are? Because otherwise some weird race-case stuff might happen. We can talk about this tomorrow, but here's what I am worried about:

Player A gets the code for his tank, and Player B's tank. Now both codes are stored on Player A's computer.

Player A runs the simulation for 1 minute to see who wins.

While A is running the sim, B changes his tank or his tanks code in such a way that, if there was to be a rematch, B would win now.

A finishes the simulation and reports that A beat B's tank.

The server get's A's tank and B's new tank (now updated, but not what A played against), and stores them saying that in the matchup, A won.

B watches the matchup with his new tank and A's tank, and B clearly wins it, leaving B very confused as to why he lost.

SecondThread: idk the best way to avoid this. We could have the client send the server the tanks that were involved (although this is ripe for cheating), or we could have the client say "I want to have a match" have the server save snapshots of the tanks, and then have the client say "get me the tanks of the match that I just wanted to have", and then have the client say "report that I won the match that you just sent me". If we are doing the later, then we would need a state for: "The client wanted the match to happen, but no one has told the server who wins it yet"

SecondThread (approved):

src/models/battleRecordModel.js
Derplime: We can hold on this and talk about it tomorrow. This was something extra that I felt could be done without much discussion since I had finished what I was supposed to do this week.

PR #121, opened 2020-03-02
Search players and its CSS and BattleArena changes integrated.

### LINKED ISSUE #106: Search Players searches a list and returns a list
Build a component that takes a list and compares the user's search keyword to that list. The result will return a new list to display users with that keyword in the name.

N/A

**Additional context**
N/A

blaiadam:
**Description**
SearchPlayers component integrated for the Battle Arena.

Resolves #106 

**Type of change**

1. Start website.
2. Go to Battle Arena.
3. Try to search for players.

**Final Checklist:**


src/battlearena/SearchPlayers.js
```diff
@@ -0,0 +1,93 @@
+//@flow strict
+
+import * as React from 'react';
+
+type Props = {|
+	playerList: Array<string>,
```
SecondThread, line 6: Doesn't matter for now, but in the future, this shouldn't be passed in as a prop, it should be called within the component. There is no reason for everyone else to have a list of all of the players, so this component should be the one making the API call.

src/battlearena/SearchPlayers.js
```diff
@@ -0,0 +1,93 @@
...
+
+	// When typing in the searchbar, update searchName.
+	handleChange = (e: string) => {
+		this.setState({
+			searchName: e
+		});
+	}
+
+	// When a user clicks search, handleSearch will activate.
+	handleSearch = () => {
```
SecondThread, line 45: Don't forget return types here.

I think it looks like: handleSearch = (): void => { ...
but I never do it like this, so I'm not sure

src/battlearena/SearchPlayers.js
```diff
@@ -0,0 +1,93 @@
...
+		super(props);
+
+		this.state = {
+			searchName: "",
+			searchList: []
+		}
+	}
+
+	// When typing in the searchbar, update searchName.
+	handleChange = (e: string) => {
```
SecondThread, line 38: Can I get a return type here please?
I think it looks like:
handleChange = (e: string): void => {
}
but I could be wrong, because I usually wrap my methods when I call them instead of when I hand them out as callbacks instead of when I define them.

src/battlearena/SearchPlayers.js
```diff
@@ -0,0 +1,93 @@
...
+
+	render(): React.Node {
+		return (
+			<div className="searchPlayers">
+				<h6>Click the Player's Name you wish to Challenge</h6>
+				<input type="text" className="inputText" onChange={(e) => this.handleChange(e.target.value)} placeholder="Search Players" />
+				<button type="button" onClick={this.handleSearch} className="btn">Search</button>
+				<div className="searchPlayersList">
+					<ul>
+						{this.state.searchList.map((name, index) =>
```
SecondThread, line 76: I like it

src/battlearena/SearchPlayers.js
```diff
@@ -0,0 +1,93 @@
...
+				<div className="searchPlayersList">
+					<ul>
+						{this.state.searchList.map((name, index) => 
+							<button 
+								key={index} 
+								type="button" 
+								className="clearbtn searchPlayerListItem" 
+								onClick={this.challengePlayer}
+							>
+							{name}
```
SecondThread, line 83: Indent another layer please

src/battlearena/SearchPlayers.js
```diff
@@ -0,0 +1,93 @@
...
+				<input type="text" className="inputText" onChange={(e) => this.handleChange(e.target.value)} placeholder="Search Players" />
+				<button type="button" onClick={this.handleSearch} className="btn">Search</button>
+				<div className="searchPlayersList">
+					<ul>
+						{this.state.searchList.map((name, index) => 
+							<button 
+								key={index} 
+								type="button" 
+								className="clearbtn searchPlayerListItem" 
+								onClick={this.challengePlayer}
```
SecondThread, line 81: You're going to need to pass in which player is going to be challenged

src/battlearena/SearchPlayers.js
```diff
@@ -0,0 +1,93 @@
...
+
+import * as React from 'react';
+
+type Props = {|
+	playerList: Array<string>,
+|};
+
+type State = {|
+	searchName: string,
+	searchList: Array<string>
```
SecondThread, line 11: These should be identified by some Player object later on, because multiple players could have the same name (maybe, I guess possibly not, but we should be able to handle it if they do), and you need to identify who you are going to challenge in the challengePlayer function somehow.

src/battlearena/SearchPlayers.js
```diff
@@ -0,0 +1,93 @@
...
+				list.push(player);
+			}
+		}
+
+		// Set the search result into the state.
+		this.setState({ searchList: list });
+	}
+
+	// Challenge the player chosen.
+	challengePlayer(): void {
```
SecondThread, line 64: Thanks for the return type :)

You're going to need to pass in which player should be challenged by some sort of unique identifier.

SecondThread (approved):
Most of it looks good for now, but you are going to need to make some changes when it comes to getting this stuff actually working. Make sure to look over my suggestions before pushing

PR #138, opened 2020-03-10
Added Challenge player functionality

blaiadam:
**Description**
Clicking quickplay or clicking on a user in the search players area will bring up a prompt on challenging the player. API and battleground implementation still needed.

Select Tank was also placed in there, however, it is not yet implemented.

**Type of change**

1. Go to battle arena page.
2. Click on quickplay or search for a user and click their name.
3. View the popup.

**Final Checklist:**

### DIFF
--- a/src/battlearena/ChallengePlayerPopup.js
+++ b/src/battlearena/ChallengePlayerPopup.js
@@ -1,0 +1,71 @@
+//@flow strict
+
+import * as React from 'react';
+import Popup from 'reactjs-popup';
+import { Link } from 'react-router-dom';
+
+type Props = {|
+	playerChallenged: string,
+	onChallengePlayer: (string) => void
+|};
+
+type State = {|
+	playerList: Array<string>,
+	playerChallengedWager: number,
+	userCurrency: number
+|};
+
+// Challenge Player Component. Creates a Popup to prevent accidental challenges.
+//
+// Prop Names:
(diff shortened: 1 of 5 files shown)


src/battlearena/ChallengePlayerPopup.js
```diff
@@ -0,0 +1,71 @@
+//@flow strict
+
+import * as React from 'react';
+import Popup from 'reactjs-popup';
+import { Link } from 'react-router-dom';
+
+type Props = {|
+	playerChallenged: string,
```
SecondThread, line 8: Can we get a player class that we pass around? We are going to need more than the players name in order to do these API calls (namely, we will need some sort if playerID), so we want that to get passed around.

src/globalComponents/SelectTank.js
```diff
@@ -0,0 +1,61 @@
+//@flow strict
+
+import * as React from 'react';
+
+type Props = {|
+	changeCurrentTank: (SyntheticInputEvent<HTMLInputElement>) => void
+|};
+
+type State = {|
+	tankList: Array<string>,
```
SecondThread, line 10: I think it would be helpful to pass an array of Tanks around instead of just their names.

SecondThread (approved):
Looks good for now, but in the future you are going to want to use some sort of class instead of just the names of these things so that we can easily pass more data around if we need to.

PR #161, opened 2020-03-13
Made Casus save and load automatically

SecondThread:
**Description**
This PR makes casus code save and load to the database. If you want to test it, you have to edit the tankID in saveCasus and loadCasus.

The reason this is rediculously annoying is that functions aren't saved in JSON. So to make sure we keep the functions attached to every Casus block in the tree, we need to 'revive' it (technical word, I didn't make it up). That is what I spent most of the time doing here and what the ungodly switch statement is for.

There really isn't a better way of doing that par than this unfortunately, since each of the different blocks needs to correspond to its own object, so that it has its own evaluate method.

**Type of change**

- Passes flow
- Saves and loads casus code. To verify, you need to
   - Create a new tank
   - Change the hardcoded id in saveCasus.js and loadCasus.js to the id of the tank you just created
   - then everything should work as you would expect

**Final Checklist:**


blaiadam (commented):
Looks good, just want Emil or Baylor to review the route and controller changes.

src/backend/controllers/tankController.js
```diff
@@ -7,82 +7,105 @@ const { validationResult } = require('express-validator');
...
+	await Tank.findById(req.params.tankId, function (err, tank) {
+		if (err){
+			console.error(err.message);
+			res.send(err);
+		}
+		// Update all fields of the tank except casus code
+		tank.tankName = req.body.tankName;
+		tank.userId = req.body.userId;
+		tank.components = req.body.components;
+		tank.isBot = req.body.isBot;
```
Derplime, line 82: Is there a reason why we are setting `isBot` here? When the document is made, the field is automatically made `false` by default. Unless the reason here is that we want to utilize this API for setting bots as well as player tanks?

src/backend/controllers/tankController.js
```diff
@@ -7,82 +7,105 @@ const { validationResult } = require('express-validator');
...
-            res.send(myUser.favoriteTankId);
-    });
+	await User.findById(req.user.id, 'favoriteTankId', (err: Error, myUser: User) => {
+		if(err) {
+			res.send(err);
+			console.log('could not find user');
+		} 
+		else
+			res.send(myUser.favoriteTankId);
+	});
```
Derplime, line 24: I should honestly do callbacks like these for my past API implementations. Probably will do that when I do my stuff in the morning tomorrow.

Derplime (approved):
Just have that one question, other than that it looks good to me.

src/backend/controllers/tankController.js
SecondThread: I don't think we really need to be. This is how it was before, but I just left it how it was before

PR #193, opened 2020-03-16
index.js now checks if a user is logged in before rendering a page

### LINKED ISSUE #122: LinkType.js not correctly implemented.
**Describe the bug**
LinkType.js is not implemented to check the names of Links.

**To Reproduce**
1. Change the <Link to="Page"> to a nonexistent page.
2. Run flow and see no errors are caught.
3. When clicking that link on the website, it will lead to a nonexistent page.

**Expected behavior**
LinkType needs to be imported to check the name of links.

**Screenshots**
N/A

**Additional context**
N/A

### LINKED ISSUE #145: No safety checks if a page is loaded before the user is logged in
We need a check to verify that the user is logged in on every page before that page is actually loaded. We assume that there is a cookie set, but if that cookie isn't set, then a whole bunch of pages will be broken. So we need to some way of checking if that cookie doesn't exist, and if not, then redirect to the login screen or something.

The best way of doing this is probably to have a verifyLogin function in its own file that gets called by every top-level container on ever front-end route. That component should verify the cookie exists (and possibly that it is valid).

**Additional context**

### LINKED ISSUE #203: Login should skip straight to main menu if logged in
**Describe the Feature**
If the user is logged in, it will require a click from the login page to move to main menu.

**To Reproduce**
1. Go to website.
2. Login.
3. Go to the login page through the URL.
4. Click login. (it shouldn't even show you the login page)

**Expected behavior**
Should skip to main menu.

**Screenshots**
N/A

**Additional context**
N/A

### LINKED ISSUE #205: Can still "login" as a deleted user.
**Describe the bug**
So this likely has to do with how we deal with cookies when you're not on the Fortuna site, or if you close your browser without logging out. Didn't try if running in incognito/private browsing would result in a different thing, but based on the nature of those modes for browsing, you wouldn't be able to reproduce this on them.

**To Reproduce**
On Fortuna site with a non-private browser (browser not in "Private-Browsing" mode) done locally
1. Register a new user & verify them so they can login.
2. Login as that user.
3. Close browser without logging out
4. Delete user from DB
5. Open browser and go back to Fortuna site and press "Login"
6. View that you login as the deleted user. You also hit a frontend error since you can't draw info from a deleted user's cookie.
7. Go back to home page and press "Login".
8. Step 6-7 repeats. Congratulations! We're stuck in a loop without being able to clear cookies unless you do it via the browser settings.

**Expected behavior**
Not sure how this can be handled, but this is just something to be aware of if anything. Maybe if something like this happens, whatever cookie was there is erased and they're redirected to the home page, or the cookie is erased and the login pop up works as normal. Both cases can just have a specific dialog prompt giving some notification of being unable to login with that account anymore.

**Screenshots**
N/A

**Additional context**
N/A

blaiadam:
**Description**
index.js now conditionally renders components based on whether a token is in cookies. Also, MainMenu is rendered instead of Login when a user is currently logged in.

We need to implement a API call that verify it is the logged in user's token, but currently it functions well.

Resolves #122 
Resolves #145 
Resolves #203 
Resolves #205 

**Type of change**

1. Go to website.
2. Type in the URL to go to MainMenu.
3. Should lead back to Login.
4. Login.
5. Try to head back to Login through the URL.
6. Should lead back to MainMenu.

**Final Checklist:**


src/globalComponents/LinkType.js
```diff
@@ -1,5 +1,5 @@
 //@flow strict
 
-type LinkType = 'Login' | 'MainMenu' | 'Marketplace' | 'Casus' | 'BattleArena' | 'Armory';
+type LinkType = 'Login' | 'MainMenu' | 'Marketplace' | 'Casus' | 'BattleArena' | 'Armory' | 'Battleground';
```
SecondThread: Thanks

SecondThread (changes requested):
If I am reading this correctly, this doesn't actually handle issue #145. You need to redirect to the login page when any page is loaded, not when any link is clicked. For example, someone might save bookmark on their browser, or click a link to fortunacombat.com/Armory. If you just typed that into the URL, it wouldn't redirect to the login, right?

The other thing you need to do, which probably requires talking to @Airor4 is verifying that the login token for that user is valid. The cookie might exist, but if you have logged in somewhere else, it wouldn't be accepted, so the webpage would still malfunction. 

We might also loose validity of a token when we have a page open but haven't refreshed it. This needs to be checked **every time we make an API call** so this might require extracting api calls into own files/helper functions.

IIRC, you can redirect to the login page just with something like this:
https://stackoverflow.com/questions/4744751/how-do-i-redirect-with-javascript

I am pretty sure we have some window.hrefs lying around in the code if you look for window.* you should be able to find some examples

blaiadam:
> IIRC, you can redirect to the login page just with something like this:
> https://stackoverflow.com/questions/4744751/how-do-i-redirect-with-javascript

I forgot you sent me some help, sorry should have reviewed this. I hope that my changes are sufficient.

src/index.js
```diff
@@ -21,14 +25,14 @@ const routing = (
...
 				<Route path="/Login" component={Login} />
-				<Route path="/MainMenu" component={MainMenu} />
-				<Route path="/Marketplace" component={Marketplace} />
-				<Route path="/Casus" component={CasusContainer} />
-				<Route path="/BattleArena" component={BattleArena} />
-				<Route path="/Armory" component={Armory} />
-				<Route path="/TrainingArena" component={TrainingArena} />
-				<Route path="/Battleground" component={Battleground} />
-				<Route path = '/ConfirmEmail/:token/:email' component= {ConfirmEmail}/>
+				<Route path="/MainMenu" component={(loggedIn) ? MainMenu : Login} />
```
SecondThread: If this works, that's really cool

src/index.js
```diff
@@ -13,6 +13,10 @@ import Armory from './armory/Armory.js';
 import TrainingArena from './trainingarena/TrainingArena.js';
 import Battleground from './battleground/Battleground.js';
 import ConfirmEmail from './login/ConfirmEmail.js';
+import Cookies from 'universal-cookie';
+
+const cookie = new Cookies();
+const loggedIn: boolean = (cookie.get('token')) ? true : false;
```
SecondThread: You also need to make sure the token is valid. If you log on with a second computer the token won't be valid, but it will exist.

SecondThread: You and baylor have to implement some sort of verifyToken api call to do this

Derplime (approved):
Verified manual URL entries to the following location as a non-logged in user redirect to Login:
MainMenu 👍 
Armory 👍 
Marketplace 👍 
Casus 👍 
Training Arena 👍 
Battleground 👍
Battle Arena 👍

src/battlearena/BattleArena.js
```diff
@@ -34,7 +41,7 @@ class BattleArena extends React.Component<Props, State> {
 				<Navbar
 					returnName="Back to Main Menu" 
 					pageName="Battle Arena" 
-					linkName="MainMenu"
+					linkName="/MainMenu"
```
SecondThread, line 44: should this be verifyLink(...)?

src/casus/CasusContainer.js
```diff
@@ -26,7 +27,7 @@ class CasusContainer extends React.Component<Props, State> {
 		return (
 			<div className="haveScorebarIfSmall">
 				<Navbar
-					linkName='Armory'
+					linkName='/Armory'
```
SecondThread, line 30: verifyLink?

src/mainmenu/MainMenu.js
```diff
@@ -4,34 +4,41 @@ import './MainMenu.css';
...
+		super();
+		verifyLogin();
+	}
+
 	render(): React.Node {
 		return (
 			<div id="Parent">
 				<Navbar 
-					linkName="Login" 
+					linkName="/Login"
```
SecondThread, line 22: verifyLink

src/marketplace/Marketplace.js
```diff
@@ -62,7 +63,7 @@ class Marketplace extends React.Component<Props, State> {
 		return (
 			<div id="Parent">
 				<Navbar 
-					linkName="MainMenu" 
+					linkName="/MainMenu"
```
SecondThread, line 66: ditto

src/trainingarena/TrainingArena.js
```diff
@@ -4,14 +4,21 @@ import './TrainingArena.css';
...
+		super();
+		verifyLogin();
+	}
+
 	render(): React.Node {
 		return (
 			<div id="Parent">
 				<Navbar 
-					linkName="BattleArena" 
+					linkName="/BattleArena"
```
SecondThread, line 21: ditto

SecondThread (approved):
Mostly looks good and unsubstantial to me, just a few missing verifyLinks

src/casus/CasusContainer.js
blaiadam: The verifyLink is done inside navbar.

PR #194, opened 2020-03-17
Created Mines and C4

### LINKED ISSUE #178: Game: Mines and C4 (medium/large)
-Create art for Mines, C4, and explosions
-Make explosion particle effects
-Make variables for dropping mines, read those variables in game loop and drop mines/C4 accordingly
-Blow up C4 when player triggers variable to
-Blow up mine when a player gets near it after a certain amount of time

This will require some sort of interface for game objects that need to be rendered and a particle system as well

SecondThread:
**Description**
Implemented Mines and C4 in the battle arena. The C4 you can place by setting a boolean to true and then blow up by setting that boolean to false. Mines you can place by changing a boolean from false to true in Casus. The current code allows you to have multiple (limited to 2) mines.

Mines work as follows:
There is a 100 frame 'arming' window during which the mine just sits there.
The mine will flash until a tank gets close, at which point it will be 'triggered'
Shortly after being triggered, the mine will explode.

Here's a demo of the mines and C4:
C4AndMineDemo

Here's the Casus code that does that (important stuff at the bottom):
image

Resolves #178 

**Type of change**

Tested with the above casus code and produced the gif as a result.

**Final Checklist:**


src/battleground/ImageDrawer.js
```diff
@@ -13,11 +13,11 @@ class ImageDrawer {
 		this.ctx=ctx;
 	}
 
-	draw(i: Image, center: Vec, width: number, angle: number): void {
+	draw(i: Image, center: Vec, width: number, angle: number, alpha: number = 1.0): void {
```
blaiadam, line 16: What is the purpose of the alpha?

src/battleground/GameObject.js
```diff
@@ -0,0 +1,28 @@
...
+//a simple class that can be implemented for objects that need to be rendered in the game
+//Override the following methods if you need them:
+//
+// update(Battleground): called once every fram
+//
+// render(ImageDrawer): draw yourself
+//
+// getRenderOrder(): returns the order that things should be rendered in
+
+class GameObject {
```
blaiadam, line 15: I like this.

blaiadam (changes requested):
Looks great, but there are two small Flow errors. The first one I believe might be because you need to use number instead of int. The second is just a missing import.

image

src/battleground/ImageDrawer.js
SecondThread: (I know you probably know already, but just in case): the alpha of an image is basically it's opaqueness. Something an alpha of 1.0 is 100% opaque and you can't see through it at all. Something with 0% alpha is completely transparent. 50% alpha (alpha = 0.5) is half-transparent.

This is really useful with smoke particles for example, so that they can fade out slowly to look like they are dispersing. If you watch the animated gif carefully you can see it.

blaiadam: Yeah I didn't know that, thank you.

blaiadam (approved):
Thanks for the fixes.

PR #210, opened 2020-03-19
Armory refactor

JorgeVidal29:
**Description**
I have done the api on a separate file and the variables into the state. I tried to do the tank as an object, however that gave me issues with dependency. I think it should be talked about in our meeting tomorrow. 

**Type of change**

**Final Checklist:**


src/armory/Armory.js
```diff
@@ -26,22 +28,22 @@ type State = {|
 	selectedCasusCode: string,
 	selectedIsBot: string,
 	userId: string,
+	//This is the array of options for each part of the tank
+	tankOptions : Array<Object>,
```
SecondThread: This isn't an array of any type of Object. Your code relies on the objects you put in this array having certain properties. No one else knows what those properties are though, so no one else can edit this code and be sure they didn't break something.

To fix this, we need to change this so it is either a class or a type.

src/armory/Armory.js
```diff
@@ -65,158 +67,197 @@ class Armory extends React.Component<Props, State> {
...
-		weaponTwoOptions = [{value: '', label: ''}];
-		scannerOneOptions= [{value: '', label: ''}];
-		scannerTwoOptions= [{value: '', label: ''}];
-		scannerThreeOptions= [{value: '', label: ''}];
-		jammerOptions = [{value: '', label: ''}];
-		treadsOptions = [{value: '', label: ''}];
-		singleUseItemsOne = [{value: '', label: ''}];
-		singleUseItemsTwo = [{value: '', label: ''}];
-		singleUseItemsThree = [{value: '', label: ''}];
+		this.setState({tankOptions: [{value: '', label: ''}]});
```
SecondThread: If I understand this code and react correctly, lines 87-98 force the entire page to rerender 12 times. You can make it only rerender once by doing:
```
this.setState({
  tankOptions: whatever tankOptionsShouldBe,
  chassisOptions: somethingElse,
  weaponOne: ...,
  et cetera: ...
});
```

SecondThread: It also cleans up the code a bit

src/armory/Armory.js
```diff
@@ -65,158 +67,197 @@ class Armory extends React.Component<Props, State> {
...
-				'Access-Control-Allow-Origin': '*',
-				'Content-Type': 'application/json',
-				'Access-Control-Allow-Credentials': 'true',
-				'x-auth-token': token
-			},
-		});
-		const body = await response.text();
-		this.setState({selectedTankId: body});
-		this.getSelectedTank();
+	getFavoriteTank () {
```
SecondThread: You need to specify the return type of this function

src/armory/Armory.js
```diff
@@ -65,158 +67,197 @@ class Armory extends React.Component<Props, State> {
...
+		const responsePromise = getFavoriteTankID();
+		responsePromise.then(
+			response => response.json().then(data => {
+				console.log(data);
```
SecondThread: If everything works correctly and is robust, then we should have debug output in production

src/armory/Armory.js
```diff
@@ -65,158 +67,197 @@ class Armory extends React.Component<Props, State> {
...
+				if (response.status !== 200) {
+					console.log(response.status);
+					console.log(data.msg);
+					console.log(data);
+					return data;
+				}
+				else {
+					let favoriteTankID = data.text();
```
SecondThread: should be const

src/armory/Armory.js
```diff
@@ -65,158 +67,197 @@ class Armory extends React.Component<Props, State> {
...
+				}
+				else {
+					let jsonObjectOfTanks = data;
+					//This will get the seleced tanks info and fill out the selected items
+					for (const tank in jsonObjectOfTanks) {
+						let obj = {};
+						obj['value'] = jsonObjectOfTanks[tank]._id;
+						obj['label'] = jsonObjectOfTanks[tank].tankName;
+						if(jsonObjectOfTanks[tank]._id === this.state.selectedTankId) {
+							this.setState({selectedTankName:jsonObjectOfTanks[tank].tankName});
```
SecondThread: Again, these 14 lines can be one setState call with a single object passed in.

src/armory/Armory.js
```diff
@@ -65,158 +67,197 @@ class Armory extends React.Component<Props, State> {
...
+					console.log(response.status);
+					console.log(data.msg);
+					console.log(data);
+					return data;
+				}
+				else {
+					let jsonObjectOfTanks = data;
+					//This will get the seleced tanks info and fill out the selected items
+					for (const tank in jsonObjectOfTanks) {
+						let obj = {};
```
SecondThread: should be const if you aren't reassigning it

src/armory/Armory.js
```diff
@@ -65,158 +67,197 @@ class Armory extends React.Component<Props, State> {
...
+						obj['value'] = jsonObjectOfTanks[tank]._id;
+						obj['label'] = jsonObjectOfTanks[tank].tankName;
```
SecondThread: The variable names 'object', 'value', and 'label' are extremely vague and don't describe what these things are at all.

I've mentioned this on two other PRs of yours:
Please don't do `jsObject['hardcodedString'] = whatever;`. That is an equivalent and harder to read version of: `jsObject.hardcodedString = whatever;`

Finally (I assume you are going to change how you do this, so it won't matter here, but in the interest of improving for the future): If for whatever reason you do actually need to make a hardcoded object with two fields, you can do it in a cleaner way with:
```
obj = {
  value: whatever,
  label: whatever
}
```

src/armory/Armory.js
```diff
@@ -65,158 +67,197 @@ class Armory extends React.Component<Props, State> {
...
-		this.setState({userId:jsonObjectOfUser._id});
-		for (const component in jsonObjectOfUser.inventory.tankComponents) {
-			const typeOfItem = getTankComponent(verifyComponent(component));
-			//This will add the chassis that the user has
-			if(typeOfItem === 'chassis' && jsonObjectOfUser.inventory.tankComponents[component] > 0) {
-				let obj = {};
-				obj['value'] = component;
-				obj['label'] = component;
-				chassisOptions.push(obj);
+	getUserInventory () {
```
SecondThread: return type

src/armory/Armory.js
```diff
@@ -65,158 +67,197 @@ class Armory extends React.Component<Props, State> {
...
+				}
+				else {
+					let jsonObjectOfUser = data;
+					for (const component in jsonObjectOfUser.inventory.tankComponents) {
+						const typeOfItem = getTankComponent(verifyComponent(component));
+						//This will add the chassis that the user has
+						if(typeOfItem === 'chassis' && jsonObjectOfUser.inventory.tankComponents[component] > 0) {
+							let obj = {};
+							obj['value'] = component;
+							obj['label'] = component;
```
SecondThread: yikes, this has all sorts of issues.
1) obj, value, and label are very undescriptive names for people who don't know what this code does.
2) You copied and pasted `jsonObjectOfUser.inventory.tankComponents[component] > 0` six times instead of just filtering out the ones that were zero ahead of time
3) all sorts of lets that should be consts

src/globalComponents/tankAPIIntegration.js
```diff
@@ -0,0 +1,45 @@
+//@flow strict
+
+import Cookies from 'universal-cookie';
+
+/*
+	This function takes no input
+	This function gets the id of the users favorite tank
+*/
+function getFavoriteTankID() {
```
SecondThread: return types

src/globalComponents/tankAPIIntegration.js
```diff
@@ -0,0 +1,45 @@
...
+			'x-auth-token': token
+		},
+	});
+	return responsePromise;
+}
+/*
+	This function takes no input
+	This function gets all of the tanks the user is associated with
+*/
+function getAllUsersTanks() {
```
SecondThread: return types

src/globalComponents/userAPIIntegration.js
```diff
@@ -0,0 +1,24 @@
+//@flow strict
+
+import Cookies from 'universal-cookie';
+
+/*
+	This function has no input
+	This function returns a json object of the user
+*/
+function getUser() {
```
SecondThread: return types

src/armory/Armory.js
```diff
@@ -65,158 +67,197 @@ class Armory extends React.Component<Props, State> {
...
+							this.state.singleUseItemsTwo.push(obj);
+							this.state.singleUseItemsThree.push(obj);
+						}
+					}
+					//set the users id
+					this.setState({userId:jsonObjectOfUser._id});
+				}
+			})
+		).catch(
+			(error) => {
```
SecondThread: unnecessary parentheses

SecondThread (changes requested):

src/armory/OptionClass.js
```diff
@@ -0,0 +1,17 @@
+//@flow strict
+
+//This is a class for our options in a select
+class OptionClass {
+
+	value: string;
+	label: string;
+
+	constructor(value: string, label: string)
+	{
```
SecondThread: please fix to match style guide

src/armory/Armory.js
```diff
@@ -64,159 +68,188 @@ class Armory extends React.Component<Props, State> {
...
+				}
+				else {
+					const favoriteTankID = data.text();
+					this.setState({selectedTankId: favoriteTankID});
+					this.getSelectedTank();
+				}
+			})
+		).catch(
+			error => {
+				console.log('No favorite Tank');
```
SecondThread, line 124: Is this true? If the response fails it is likely due to a connection error, not something being wrong on the server, right?

src/armory/Armory.js
```diff
@@ -64,159 +68,188 @@ class Armory extends React.Component<Props, State> {
...
+					console.log(data);
+					return data;
+				}
+				else {
+					const jsonObjectOfTanks = data;
+					//This will get the seleced tanks info and fill out the selected items
+					for (const tank in jsonObjectOfTanks) {
+						//Select fields is an object of the fields that we need to construct our selects in the render
+						const tankOption = new OptionClass(jsonObjectOfTanks[tank]._id, jsonObjectOfTanks[tank].tankName);
+						if(jsonObjectOfTanks[tank]._id === this.state.selectedTankId) {
```
SecondThread, line 150: This should be moved to a tankLoader class, but we can do that later

SecondThread (approved):
Much better, thanks for the changes. One comment to fix before pushing, but looks good other than that.

src/armory/Armory.js
JorgeVidal29: Just to have an answer on this, it happens because if the user doesn't have a favorite tank the api returns nothing so the responsePromise.then gives a end of json object error.

PR #219, opened 2020-03-20
Remove all js cycles except for 4 in Casus code taht would be annoying and aren't causing problems

### LINKED ISSUE #218: All sorts of JS Cycles that cause some imports to break everything
**Describe the bug**
We currently have the following 10 JS import cycles:
image
You can see these by running npm run cycle (uses madge interally) in the removeJSCycles branch.

This is a serious problem because when we import tank from Armory.js, it causes some issue with webpack for no apparent reason (although it is likely related to the fact that we have cycles). This issue is to remove enough cycles so that, at the very least, we can import Tank.js in the armory, but hopefully remove as many as can be easily removed without an insane amount of work

**To Reproduce**
Checkout to current version of master.
Add the following line below the list of imports to Armory.js:
'import Tank from '../tanks/Tank.js'
Then try to run npm start and wait for the page to load
You get this error:
image
...although page should load correctly.

**Expected result**
The page loads fine without complaining

SecondThread:
**Description**
Removed all circular dependencies except for 4 which would require a solid amount of refactoring, and are outside of the scope of why we are removing cycles anyway:
image

Woohoo!

Resolves #218 

**Type of change**

Passes flow:
image

The page now loads when you import Tank.js from armory:
image

**Final Checklist:**


blaiadam (approved):
Looks good. Thanks for the fix, now we won't be arrested

PR #225, opened 2020-03-21
Buying and selling tanks

### LINKED ISSUE #216: Buying and Selling Tanks
The buying and selling of tanks from one user to another in the marketplace.

JorgeVidal29:
**Description**
Allows users to buy and sell tanks

Resolves #216 

**Type of change**

- [x ] New feature (non-breaking change which adds functionality)

**Final Checklist:**

- [x ] My code follows the style guidelines of this project
- [ x] I have performed a self-review of my own code
- [ x] I have commented my code, particularly in hard-to-understand areas
- [ x] My changes generate no new warnings
- [x ] Pulled updated master branch (if changed since branch creation) and corrected any conflicts, if any occur.


Derplime (approved):
LGTM

Derplime:
Do note when I do my PR for the issue about getting tank names for sale posts, you might need to update how you get those sales.

src/marketplace/ListingsView.js
```diff
@@ -65,15 +65,36 @@ class ListingsView extends React.Component<Props, State> {
...
-				const sellingObject = new SaleObject(
-					jsonObjectOfSells[sale].itemId,
-					jsonObjectOfSells[sale].salePrice,
-					jsonObjectOfSells[sale].amount,
-					jsonObjectOfSells[sale].sellerId,
-					jsonObjectOfSells[sale]._id);
-				itemsForSaleArray.push(sellingObject);
+			//If we have tanks we need to process those sells different
+			if(this.props.sellerType === 'tanks')
+			{
```
SecondThread: Make sure you match the style guide before making a PR

SecondThread (changes requested):

src/marketplace/ListingsView.js
```diff
@@ -65,15 +65,36 @@ class ListingsView extends React.Component<Props, State> {
...
+				//if this isn't a component it must be a tank so we can process it here
+				if(getTankComponent(jsonObjectOfSells[sale].itemId) === undefined)
```
SecondThread: == null preferred. Handles both null and undefined

src/marketplace/ListingsView.js
```diff
@@ -65,15 +65,36 @@ class ListingsView extends React.Component<Props, State> {
...
-			const typeOfItem = getTankComponent(verifyComponent(jsonObjectOfSells[sale].itemId));
-			if(typeOfItem === this.props.sellerType) {
-				const sellingObject = new SaleObject(
-					jsonObjectOfSells[sale].itemId,
-					jsonObjectOfSells[sale].salePrice,
-					jsonObjectOfSells[sale].amount,
-					jsonObjectOfSells[sale].sellerId,
-					jsonObjectOfSells[sale]._id);
-				itemsForSaleArray.push(sellingObject);
+			//If we have tanks we need to process those sells different
```
SecondThread: *differently
*sales, not sells

src/marketplace/MakeATankSaleView.js
```diff
@@ -0,0 +1,118 @@
...
+import {makeASale} from './marketPlaceAPIConnections.js';
+
+type Props = {||}; 
+type State = {|
+	userId: string,
+	salePrice: number,
+	tankBeingSoldId: string,
+	itemAmount: number,
+	tanksToSell: Array<OptionClass>,
+|};
```
SecondThread, line 17: new line between types and class declaration please

src/marketplace/MakeATankSaleView.js
```diff
@@ -0,0 +1,118 @@
...
+		const responsePromise = makeASale(this.state.userId, this.state.salePrice, this.state.tankBeingSoldId, 'tank', 1);
+		responsePromise.then(
+			response => response.json().then(data => {
+				console.log(data);
+				//lets get the new tanks that we have since we lost the current one
+				this.getAllUsersTanksForSell();
+				//set the new selling price to zero 
+				this.setState({salePrice: 0});
+				}
+			)
```
SecondThread: }) should be on the same line so the indentation matches up

SecondThread: Also, you are dealing with the backend here which could fail, so you need to handle what happens if it does.

src/marketplace/MakeATankSaleView.js
```diff
@@ -0,0 +1,118 @@
...
+			error => {
+				console.log('Couldnt connect to server!');
+				console.log(error);
+			}
+		);
+	};
+
+	//This gets all of the users tanks and then adds them to the dropdown
+	getAllUsersTanksForSell() : void {
+		const responsePromise = getAllUsersTanks();
```
SecondThread, line 62: You should tank to Adam about this since you guys will both be touching this, but getAllUserTanks should return an array of Tank objects. Things get very messy if you are touching how the backend stores this stuff as JSON in multiple places because that makes it so that if we ever change how the backend stores this stuff, we have to change the entire front end as well, which is terrible design.

src/marketplace/MakeATankSaleView.js
```diff
@@ -0,0 +1,118 @@
...
+	};
+
+	handleChangeInSaleItem = ({ target }:{target:HTMLInputElement }) => {this.setState({tankBeingSoldId: target.value});}
+	handleChangeInSalePrice = ({ target }:{target:HTMLInputElement }) => {this.setState({salePrice: parseInt(target.value)});}
+	
+	render(): React.Node  { 
+		return (
+			<div id="Parent">
+				<label>Select a tank to Sell</label>
+				<select className="tankForSell" onChange={this.handleChangeInSaleItem}>{this.state.tanksToSell.map(({ value, label }, index) => <option key={index}  value={value}>{label}</option>)}</select>
```
SecondThread: It would be much more helpful to be processing a tank object here. That way your state represents the data that is being rendered, not the way in which that data is being rendered.

SecondThread (changes requested):

Derplime (changes requested):
I merged #232 into `master`, so this may have to be updated to use the new API I made to handle getting the list of tanks for sale so you can access the tanks' names.

src/marketplace/MakeATankSaleView.js
JorgeVidal29: Ok will talk to him about this

src/marketplace/MakeATankSaleView.js
JorgeVidal29: That would be a backend question I think. That goes into the talk about what we were talking about in the meeting on 3/21.

Derplime (changes requested):
When I tested this, when the tank was bought, it deducted **twice** the sales amount and gave it to the seller. The API hasn't changed for this, so I'm not sure what's going on.

Also for here: 
image

While you made the Category selection not clickable, is it also possible to change the color of the selection itself or something to make it stand out more as just a column header? (Optional really, the real issue is the first thing)

Other than that, everything worked well (up to the double money transaction) and looks great.

src/marketplace/ListingsView.js
```diff
@@ -52,33 +52,119 @@ class ListingsView extends React.Component<Props, State> {
 		this.setState({userId:jsonObjectOfUser._id});
 	};
 
+
+	//This function directs the view to the proper function
+	//If we are in the tank view it directs it to getMarketSalesForTanks()
+	//Else it directs it to getMarketSalesForComponents()
+	directSaleToProperFunction() : void {
+		if(this.props.sellerType === 'tanks')
+		{
```
SecondThread: please fix to match style guide

src/marketplace/ListingsView.js
```diff
@@ -52,33 +52,119 @@ class ListingsView extends React.Component<Props, State> {
...
+			this.getMarketSalesForTanks();
+		}
+		else
+		{
```
SecondThread: ditto

src/marketplace/ListingsView.js
```diff
@@ -52,33 +52,119 @@ class ListingsView extends React.Component<Props, State> {
...
+		responsePromise.then(
+			response => response.json().then(data => {
+				if (response.status !== 200) {
+					console.log(response.status);
+					console.log(data.msg);
+					console.log(data);
+					return data;
+				}
+				else {
+					const jsonObjectOfSells = data;
```
SecondThread, line 89: nit: the itemsForSale array should probably be in this scope

src/marketplace/ListingsView.js
```diff
@@ -52,33 +52,119 @@ class ListingsView extends React.Component<Props, State> {
...
+					console.log(data);
```
SecondThread: no misc. logs in prod please

src/marketplace/MakeATankSaleView.js
SecondThread: I'm going to stick with this one. Your state should represent the information you are rendering. What you have here doesn't do that. What you are storing in your state _is the way you render that tank_ rather than the tank you are rendering. IIRC, Adam's latest PR handles converting the backend version of the tank into the front-end version.

src/marketplace/MakeATankSaleView.js
SecondThread: Re-requesting that we don't write duplicate code to process getAllTanks

src/marketplace/MakeATankSaleView.js
```diff
@@ -0,0 +1,122 @@
...
+import {getUser} from '../globalComponents/userAPIIntegration.js';
+import {makeASale} from './marketPlaceAPIConnections.js';
+
+type Props = {||}; 
+type State = {|
+	userId: string,
+	salePrice: number,
+	tankBeingSoldId: string,
+	itemAmount: number,
+	tanksToSell: Array<OptionClass>,
```
SecondThread: Please make this an array of tanks, instead of an array the things that should be rendered. Like I mentioned on my previous PR review, your state should represent the data you are rendering, not the way you render it. I'm happy to talk with you about why this is better design if you would like.

SecondThread: It will probably be helpful to talk to Adam since he wrote a bunch of code to do this.

SecondThread (changes requested):

Derplime:
Fixed the issue with the sales costing users double for some reason. So the (err: Error) callbacks that we are doing in the mongoose calls are actually calling the function multiple times.

Quote from [this issue](https://github.com/Automattic/mongoose/issues/6271)
>When you pass a callback to your query, you execute the query. Then, when you await on the query, you execute it again, so you're actually executing the query twice by passing that unnecessary callback.

src/marketplace/ListingsView.js
```diff
@@ -32,7 +32,7 @@ class ListingsView extends React.Component<Props, State> {
 
 	//When sellerType is updated we need to get the new sells
 	componentDidUpdate(prevProps:Props) {
```
SecondThread: return type void

src/marketplace/ListingsView.js
```diff
@@ -52,33 +52,116 @@ class ListingsView extends React.Component<Props, State> {
...
+		responsePromise.then(
+			response => response.json().then(data => {
+				if (response.status !== 200) {
+					console.log(response.status);
+					console.log(data.msg);
+					console.log(data);
+					return data;
+				}
+				else {
+					const jsonObjectOfSells = data;
```
SecondThread, line 138: nit: (same thing) itemsForSaleArray should probably be in this scope

src/marketplace/ListingsView.js
```diff
@@ -52,33 +52,116 @@ class ListingsView extends React.Component<Props, State> {
...
+					for (const sale in jsonObjectOfSells) {
+						//if this isn't a component it must be a tank so we can process it here
+						if(getComponentType(jsonObjectOfSells[sale].itemId._id) == null) {
+							console.log(jsonObjectOfSells[sale]);
```
SecondThread: No misc. print statements in prod please

src/marketplace/ListingsView.js
```diff
@@ -32,7 +32,7 @@ class ListingsView extends React.Component<Props, State> {
```
SecondThread, line 32: ^The class `SaleObject` should be defined in the file `SaleObject.js`, not `saleObjectClass.js`

src/marketplace/ListingsView.js
```diff
@@ -52,33 +52,116 @@ class ListingsView extends React.Component<Props, State> {
 		this.setState({userId:jsonObjectOfUser._id});
 	};
 
+
+	//This function directs the view to the proper function
+	//If we are in the tank view it directs it to getMarketSalesForTanks()
+	//Else it directs it to getMarketSalesForComponents()
+	directSaleToProperFunction() : void {
+		if(this.props.sellerType === 'tanks') {
```
SecondThread, line 60: Wait, what? `sellerType` is of type ComponentType, which is: `'chassis' | 'weapon' | 'scanner' | 'scannerAddon' | 'jammer' | 'treads' | 'item'`. So if this is true, we have a very big problem that is causing your props to not be the type the code thinks it is.

src/marketplace/MakeATankSaleView.js
SecondThread: Is this not a perfect example of shared functionality? All this method is doing is getting all tanks for a user as Tank objects, no? So this method shouldn't just call some method that gets the backend tanks and the armory should call the same method. Otherwise we have duplicate code here, and then people who are debugging this code later would get very confused as to why sometimes Tank objects have all of the data, and sometimes they only have random fields like 'tankName' and '_id'. 

If I am missing something and you actually think we need to do this a second time, we can talk about it, but it doesn't look like you addressed this at all

src/marketplace/Marketplace.js
```diff
@@ -49,12 +50,18 @@ class Marketplace extends React.Component<Props, State> {
 			case 'ACCESSORIES':
 				partView=(<ListingsView sellerType='item'></ListingsView>);
 				break;
+			case 'TANKS':
+				partView=(<ListingsView sellerType='tanks'></ListingsView>);
```
SecondThread, line 54: I don't think that is allowed. `sellerType` doesn't include 'tanks', does it?

SecondThread (changes requested):
Doesn't pass flow:
image

blaiadam:
Fails flow
image

src/marketplace/MakeATankSaleView.js
blaiadam: For some reason, returning tanks from the tankAPIIntegration.js breaks down a lot of what we have been doing. I think this is a problem that will take some changes that I am not aware on how to do. I suggest leaving it for now, but if you want to discuss this with me, feel free to message me.

src/marketplace/ListingsView.js
```diff
@@ -52,33 +52,115 @@ class ListingsView extends React.Component<Props, State> {
...
+				console.log(error);
+				return error;
+			}
+		); 
 	}
-	
+
 	//This creates a card for every sale
 	createCards = () => {
 		const cards = []
```
blaiadam, line 166: You need to change the components being listed to be in title case. It is difficult to read otherwise (shortRangeJammer vs Short Range Jammer). There is a function in armory using it.

blaiadam (changes requested):
When making a listing, I have no way of knowing how many items I have. You need to show that to the user so that they can be able to specify how much. Also there is no feedback when I choose too much. Check out other comments.

src/armory/Armory.js
```diff
@@ -14,6 +14,7 @@ import { getInventory, getComponentPoints } from './GetInventoryInfo.js';
 import { getUser } from '../globalComponents/userAPIIntegration.js';
 import { getAllUsersTanks } from '../globalComponents/tankAPIIntegration.js';
 import { getTank, getEmptyCasusCode } from '../tanks/TankLoader.js';
+import { toTitleCase } from '../globalComponents/Utility.js';
```
blaiadam, line 17: Thank you for doing that.

blaiadam (changes requested):
Looks good, just one change. We can't let a user sell their last tank. User's must have at least one tank.

blaiadam (approved):
Make the if statement you put in match the style we use and you can merge.

Derplime (approved):

JorgeVidal29:
> When making a listing, I have no way of knowing how many items I have. You need to show that to the user so that they can be able to specify how much. Also there is no feedback when I choose too much. Check out other comments.

Just to have an answer to this written down, this is a change for buying and selling of tanks not any items.

SecondThread (approved):
Mostly looks good to me. We can talk about writing a method to get all tanks during the meeting I guess

Issue #157, opened 2020-03-13
No response from API call when the login token isn't provided


SecondThread:
**Describe the bug**
Everything works fine when the x-auth-token is provided in the header. When the x-auth-token isn't (and possibly when it is but is incorrect), there is just no response to the front end. This makes error handling really hard.

**To Reproduce**
1. Call saveCasus for example but comment out the part about the x-auth-token

**Expected behavior**
Some sort of error or something saying that the auth token is invalid.

**Screenshots**

**Additional context**

Airor4:
Is this the /casusUpdate api call you're talking about because every api call with auth middleware i see has token error handling?

SecondThread:
TODO: talk with baylor after the meeting

Airor4:
That particular route didn't have authentication on it. It has since been added

PR #223, opened 2020-03-21
Api runthrough

Derplime:
**Description**
Did a run-through of the APIs to clean them up.
Any additions were for error handling or consistency of returning responses where necessary.
Overall none of these updates affects functionality.
See #213 for a general list of what I ran through. Not everything is there, but it's a example list of what I looked into.

Resolves #213 

**Type of change**
Check options that are relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Fix or feature that would cause existing functionality to not work as expected
- [ ] This change requires a update in the design document

**Steps to Verify Functionality**
* Verify any desired API still works after these changes.

Passes flow:
image

**Final Checklist:**
Go down the list and check them off.

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the design document (if applicable)
- [x] My changes generate no new warnings
- [x] Pulled updated master branch (if changed since branch creation) and corrected any conflicts, if any occur.

PR #231, opened 2020-03-22
Api jwt verification

### LINKED ISSUE #173: is token valid api call
Check if token is valid for current session based on login and time restrictions.

Airor4:
**Description**
Created a route to specifically verify jwt tokens and added auth to updateCassus route.
Resolves #173 #157 
**Type of change**

- [X] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)

Call the /verify route on Postman with a x-auth-token header received from login:
* Header key: `x-auth-token`
* Header value: Token received on login. No quotes.

See if it returns a success response. 
Also call the /updateCasus route and see if the auth gives feedback.

**Final Checklist:**

- [X] I have made corresponding changes to the design document (if applicable)
- [X] My changes generate no new warnings
- [X] Pulled updated master branch (if changed since branch creation) and corrected any conflicts, if any occur.


src/backend/controllers/tankController.js
```diff
@@ -1,68 +1,64 @@
...
 const { validationResult } = require('express-validator');
 
-exports.getFavorite = async (req: Request, res: Response) => {
+// Flow type import
+import type {
+	$Request,
+	$Response,
+	NextFunction,
+	Middleware,
+} from 'express';
```
Derplime: So is there a reason why it was moved here rather just just appended to the previous implementation?

Also is there a reason why you need `NextFunction` and `Middleware` if these imports are not used at all?

src/backend/controllers/tankController.js
```diff
@@ -1,68 +1,64 @@
...
+
+exports.getFavorite = async (req: $Request, res: $Response) => {
```
Derplime: I guess I might be out of the loop when it comes to it, but what is the significance of `$Request/$Response` vs `Request/Response`? If this is about defining types of the arguments for Flow, the latter ones are listed rather than the former.

Might be out of date, but I use [this](https://www.saltycrane.com/cheat-sheets/flow-type/latest/) for the Flow types

src/backend/controllers/tankController.js
```diff
@@ -1,68 +1,64 @@
...
-		}
 	});
 }
 
-exports.favoriteTank = async (req: Request, res: Response) => {
+exports.favoriteTank = async (req: $Request, res: $Response) => {
 	// the 'new' option means return the document after it has been updated
-	await User.findOneAndUpdate( { _id: req.user.id }, {favoriteTankId : req.body.favoriteTankId}, {'new':true }, (err: Error, result: User) => {
+	// the 'useFindAndModify' is just so it doesnt throw a deprecation warning
+	await User.findOneAndUpdate( { _id: req.user.id }, {favoriteTank : req.body.favoriteTankId}, {'new':true, 'useFindAndModify':false}, (err: Error, result: User) => {
```
Derplime: I removed that `useFindAndModify` for a reason. It's redundant since it's used as a general option in `server.js`
image

Unless this call still complains about this option despite it being in the options in the server.

src/backend/controllers/tankController.js
```diff
@@ -1,68 +1,64 @@
...
 	const tank = new Tank();
 	tank.userId = req.user.id;
 	tank.tankName = req.body.tankName;
-	await tank.save((err: Error) => {
-		if (err) {
-			res.send(err);
-			console.error(err.message);
-		}
-	});
+	await tank.save();
```
Derplime: Add error handling for the save.

src/backend/controllers/tankController.js
```diff
@@ -84,18 +80,12 @@ exports.tankUpdate = async (req: Request, res: Response) => {
 		tank.userId = req.body.userId;
 		tank.components = req.body.components;
 		tank.isBot = req.body.isBot;
-		tank.save((err: Error) => {
-			if (err) {
-				console.error(err.message);
-				res.send(err);
-			}
-		});
+		tank.save();
```
Derplime: Error handling for the save.

src/backend/controllers/tankController.js
```diff
@@ -107,21 +97,15 @@ exports.casusUpdate = async (req: Request, res: Response) => {
...
 		}
 		// Update the casus code
 		tank.casusCode = req.body.casusCode;
-		tank.save((err: Error) => {
-			if (err) {
-				console.error(err.message);
-				res.send(err);
-			}
-		});
+		tank.save();
```
Derplime: Error handling for the save.

src/backend/controllers/tankController.js
```diff
@@ -107,21 +97,15 @@ exports.casusUpdate = async (req: Request, res: Response) => {
 			.json({ errors: errors.array() });
 	}
 
-	await Tank.findById(req.params.tankId, function (err, tank) {
-		if (err){
+	await Tank.findById(req.params.tankId, function (err, tank){ 
+		if(err){
```
Derplime: Not sure why there were changes here. And even so, as picky as it is, spacing between things are off. (optionalish)

src/backend/controllers/verifyController.js
```diff
@@ -0,0 +1,19 @@
...
+
+exports.verify = async (req, res) => {
+    const token = req.header('x-auth-token');
+    if(!token) {
+		return res.status(401).json({ msg: 'No token, authorization denied' });
+    }
+
+    try{
+        jwt.verify(token, jwtSecret);
+        res.status(200).json({ msg: 'Token valid!'});
```
Derplime: Return result Ex: `return res.status.......`.

src/backend/controllers/verifyController.js
```diff
@@ -0,0 +1,19 @@
...
+    }catch(err){
+        res.status(401).json({ msg: 'Token is not valid' });
```
Derplime: Same as above

src/backend/routes/userRoutes.js
```diff
@@ -28,7 +28,7 @@ router.post('/registerUser', [
 
 // Login a User
 // Route call: /login
-// Req must include email and password in body
+// Req must include userName and password in body
```
Derplime, line 31: Nice catch, forgot to update that.

src/backend/controllers/verifyController.js
```diff
@@ -0,0 +1,19 @@
+const jwt = require('jsonwebtoken');
```
Derplime, line 3: Flow strict. Also header comments similar to other controllers.

src/backend/routes/verifyRoute.js
```diff
@@ -0,0 +1,9 @@
+const express = require('express');
```
Derplime, line 7: Flow strict. Also header comments similar to other route files.

Derplime (changes requested):
Left a good amount of things to address. Some of the changes essentially reverted the ones I made in PR #223. I'd like your opinion on these, so do try to elaborate on each thing as much as possible.

src/backend/controllers/tankController.js
```diff
@@ -1,68 +1,64 @@
...
-	await User.findOneAndUpdate( { _id: req.user.id }, {favoriteTankId : req.body.favoriteTankId}, {'new':true }, (err: Error, result: User) => {
+	// the 'useFindAndModify' is just so it doesnt throw a deprecation warning
+	await User.findOneAndUpdate( { _id: req.user.id }, {favoriteTank : req.body.favoriteTankId}, {'new':true, 'useFindAndModify':false}, (err: Error, result: User) => {
 		if(err) {
 			res.status(500).send(err);
 			console.error(err);
 		}
-		else {
-			res.send(result.favoriteTankId);
-			console.log('Updated favorite tank.');
```
Derplime, line 33: I guess this is more of a question of whether it's right to make logs for successful API calls.

Some controllers are using `return res.status(200).json({ msg: 'Success message here' });` or
```
console.log('Success message here');
return res.status(200).json(dbObject);
```
What makes these an exception to not including some log going out to confirm a route worked as it should?

src/backend/controllers/tankController.js
```diff
@@ -1,68 +1,64 @@
...
+exports.userTanks = async (req: $Request, res: $Response) => {
 	await Tank.find({ userId: req.user.id }, (err: Error, tanks: Array<Tank>) => {
 		if(err) {
 			res.send(err);
 			console.error(err.message);
 		}
-		else {
+		else{
 			res.send(tanks);
-			console.log('Retrieved user\'s tanks.');
```
Derplime, line 46: See above question regarding having logs with successful calls.

src/backend/controllers/tankController.js
```diff
@@ -1,68 +1,64 @@
...
 	tank.tankName = req.body.tankName;
-	await tank.save((err: Error) => {
-		if (err) {
-			res.send(err);
-			console.error(err.message);
-		}
-	});
+	await tank.save();
 	res.status(201).send(tank);
-	console.log('Successfully created a tank.')
```
Derplime, line 62: See above question regarding having logs with successful calls.

src/backend/controllers/tankController.js
```diff
@@ -84,18 +80,12 @@ exports.tankUpdate = async (req: Request, res: Response) => {
...
 		tank.isBot = req.body.isBot;
-		tank.save((err: Error) => {
-			if (err) {
-				console.error(err.message);
-				res.send(err);
-			}
-		});
+		tank.save();
 		res.status(200).send(tank);
-		console.log('Successfully updated tank.')
```
Derplime, line 94: See above question regarding having logs with successful calls.

src/backend/controllers/tankController.js
```diff
@@ -107,21 +97,15 @@ exports.casusUpdate = async (req: Request, res: Response) => {
...
 		tank.casusCode = req.body.casusCode;
-		tank.save((err: Error) => {
-			if (err) {
-				console.error(err.message);
-				res.send(err);
-			}
-		});
+		tank.save();
 		res.status(200).send(tank);
-		console.log('Successfully updated tank Casus code.')
```
Derplime, line 124: See above question regarding having logs with successful calls.

src/backend/controllers/tankController.js
SecondThread: +1 to using Request/Response instead of $Request/$Response unless there is a very compelling reason not to

src/backend/controllers/tankController.js
```diff
@@ -1,68 +1,64 @@
...
+} from 'express';
+
+exports.getFavorite = async (req: $Request, res: $Response) => {
 	await User.findById(req.user.id, 'favoriteTankId', (err: Error, myUser: User) => {
 		if(err) {
 			res.send(err);
 			console.log('could not find user');
 		} 
-		else {
+		else
```
SecondThread: Please keep the curly braces to match our style guide.

(Personally, I don't like the unnecessary ones either, but stuff should be consistent so that it is easier to read)

src/backend/controllers/tankController.js
```diff
@@ -1,68 +1,64 @@
...
 			res.send(myUser.favoriteTankId);
-			console.log('Retrieved Favorite Tank.');
-		}
 	});
 }
 
-exports.favoriteTank = async (req: Request, res: Response) => {
+exports.favoriteTank = async (req: $Request, res: $Response) => {
```
SecondThread: Do we have a compelling reason for this?

src/backend/controllers/tankController.js
SecondThread: I have no strong feelings either way. I'm a little bit partial to successful console.logs() because that way when something inevitably _doesn't_ work, we can see that it is the front-end's fault/a connection issue more easily.

We should probably pick something and be consistent with it though.

src/backend/controllers/tankController.js
SecondThread: idk, not really optional in my book. It isn't hard to look through your own stuff before a PR, so make it match the style guide please.

src/backend/controllers/verifyController.js
SecondThread: **FLOW**

src/backend/controllers/verifyController.js
```diff
@@ -0,0 +1,19 @@
...
+exports.verify = async (req, res) => {
+    const token = req.header('x-auth-token');
+    if(!token) {
+		return res.status(401).json({ msg: 'No token, authorization denied' });
+    }
+
+    try{
+        jwt.verify(token, jwtSecret);
+        res.status(200).json({ msg: 'Token valid!'});
+    }catch(err){
```
SecondThread: space after } please

src/backend/routes/verifyRoute.js
SecondThread: +1

src/index.js
```diff
@@ -8,7 +8,7 @@ import Login from './login/Login.js';
 import MainMenu from './mainmenu/MainMenu.js';
 import Marketplace from './marketplace/Marketplace.js';
 import CasusContainer from './casus/CasusContainer.js';
-import BattleArena from './battlearena/BattleArena.js';
+import BattleArena from './BattleArena/BattleArena.js';
```
SecondThread: pretty sure it should still be battlearena unless you moved something?

SecondThread (dismissed):

src/backend/controllers/tankController.js
Airor4: Going to remove it for now because I want everything to be uniform. Will shape it up with console logs when I have the time.

src/backend/controllers/tankController.js
```diff
@@ -1,49 +1,66 @@
...
-// Flow type import
-import type { Request, Response } from 'express';
-
 const User = require('../../models/userModel');
 const Tank = require('../../models/tankModel');
 
 const { validationResult } = require('express-validator');
 
+// Flow type import
+import type { Request, Response } from 'express';
```
Derplime, line 9: If this was unchanged this should have stayed where it originally was at the top (it's mostly because of syntax reasons with putting your imports before anything else, but functionally it works still), but it's fine.

Derplime (changes requested):
Was able to have the /verify route itself work (via updated verification instructions), but get back to me on getting the other route to work to test correctly.

UPDATE:

Got updateCasus route to work, but the tank returned is not changed at all in the DB (confirmed in MongoDB Compass), however you do have the `new: true` option there.
image
Does the field not accept some input like this?

Derplime (approved):
Thanks for the catch on that. It worked fine now on my end.
image

You'll have to fix the conflict that's there before being able to merge though.

PR #244, opened 2020-03-24
Guns and Roses: Massive Art Dump. Guns PR part 1

SecondThread:
**Description**
Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

**Type of change**

Renders tanks with new guns normally. Not much code changed, just a massive amount of art here. (bullets aren't actually created yet). Guns slow you down

**Final Checklist:**


src/tanks/Tank.js
```diff
@@ -154,9 +153,6 @@ class Tank extends GameObject {
 		//end of movement stuff
 		
 		//gun stuff
-		const turretDirection = this._getDouble(TURRET_DIRECTION_VAR_NAME);
-		this.mainGun.setTargetGunAngle(turretDirection);
-		this.mainGun.onUpdate();
 		//end of gun stuff
```
Derplime, line 156: Reason for this removal is since you're moving all turret direction to the Gun.js correct? Just for clarifying.

Derplime (approved):
Had a question of clarification below, though I probably got the answer for it after a second pass at things. Other than that LGTM

src/tanks/Tank.js
SecondThread: Yeah, all components have their own update call now (called like 20 lines down or so), so the turret can just use that method instead of being directly told what to do from the tank.

And it is now controlling itself, correct.

PR #246, opened 2020-03-24
Made Secondary weapons load and render correctly

### LINKED ISSUE #201: Game: Handle secondary weapons mounted to chassis (easy)
In theory we can have two weapons per tank, but in practice only one is supported in the game at the moment.

A quick fix for the moment would be to just mount the secondary weapon to the chassis (render it before the chassis), which would probably look pretty meh. A better solution would be to have a second image for every gun for when it is attached to the chassis, and then for those we can render those on top of the chassis, but keep them fixed to the chassis's rotation

SecondThread:
**Description**
Secondary weapons are now controllable with a separate Casus variable and are mounted to the bottom of the chassis. They still affect movement speed to discourage users from just leaving them on forever, and look like this:
SecondaryWeapons2

I looked through all the weapons being used as secondaries, and they all look okay. Some aren't great, but they are identifiable, and we are really starting to run out of places to put things lol so I don't know how much better it can get.

Resolves #201 

**Type of change**

**Final Checklist:**


src/battleground/ImageLoader.js
```diff
@@ -158,6 +158,9 @@ function _anotherImageLoaded(): void {
 }
 
 function getImage(name: ImageName): Image {
+	if (name==null) {
+		throw new Error('asdfasdf');
```
Derplime: That supposed to be that text? lol

src/tanks/Tank.js
```diff
@@ -84,9 +86,10 @@ class Tank extends GameObject {
 		this.chassis = chassis;
 		this.treads = treads;
 		this.mainGun = mainGun;
+		this.secondaryGun = secondaryGun;
 		this.scanner = scanner;
 		this.jammer = jammer;
-		this.parts = [chassis, treads, mainGun, scanner, jammer];
+		this.parts = [chassis, treads, mainGun, secondaryGun, scanner, jammer];
```
Derplime, line 92: Is this loading something different from what Adam had here?
>Parts will be an array of 11 values:
[0] = Chassis
[1] = Main Gun, [2] = Secondary Gun
[3] = Scanner, [4] = Scanner Addon 1, [5] = Scanner Addon 2
[6] = Jammer
[7] = Treads
[8] = Item 1, [9] = Item 2, [10] = Item 3

Derplime (changes requested):
It looks good to me, though I had two questions that I'd like you to address and if necessary, make changes.

src/battleground/gameobjects/Bullet.js
```diff
@@ -0,0 +1,205 @@
...
+		speed: 3,
+		width: 1,
+		lifetime: 0,
+	}
+};
+
+class Bullet extends GameObject {
+
+	rotation: number;
+	tankToIgnore: Tank;
```
blaiadam, line 87: Is this supposed to be the user's tank that is firing the bullet?

src/tanks/Gun.js
```diff
@@ -6,20 +6,38 @@ import {verifyBoolean, verifyDouble} from '../casus/interpreter/Value.js';
...
+import type {BulletType} from '../battleground/gameobjects/Bullet.js';
 
 const LERP_PERCENT=0.2;
 const GUN_CENTER_TO_TANK_CENTER=2;
 const GUN_CENTER_TO_GUN_ROT=2;
 const TAU=Math.PI*2;
 const FIRING_SLOWDOWN=0.6;
 
+//
+// Which guns are which types:
```
blaiadam, line 27: Thank you for this comment. I'm gonna change so the type is recognized in the constructor based on the name instead of having it passed as a parameter.

src/battleground/gameobjects/Bullet.js
SecondThread: yeah

SecondThread: (assuming you can't shoot yourself, which I thought was a fair assumption to make)

src/battleground/ImageLoader.js
SecondThread: oh yeah, I guess I probably don't need that lol

src/tanks/Tank.js
SecondThread: I think Adam's array is the way things are stored on the back end. This is just the list of all of the tankparts used in the game. So I think they are different, yeah

src/tanks/Gun.js
SecondThread: Sounds good to me

src/tanks/Tank.js
Derplime: Alright that's all I was concerned about. Thanks for clarifying.

Derplime (approved):
Thanks for addressing my questions 👍

PR #247, opened 2020-03-25
New tanks for newbs

### LINKED ISSUE #243: When a user is created, they should have a free tank
Allowing a user to have no tanks can cause issues. This needs to be discussed before the CurrentTank PR is pushed to master.

Add handling on the frontend for when a user has no tanks.

**Additional context**
If this is made the case, then we can't let user's delete their last tank.

Derplime:
**Description**
Updated registerUser API to give newly registered users an initial tank.

Resolves #243 

**Type of change**

1. Register a new user.
2. Verify on MongoDB Compass that an initial tank has been made owned by the user's id.

**Final Checklist:**

Passes Flow:
image

### DIFF
--- a/src/backend/controllers/userController.js
+++ b/src/backend/controllers/userController.js
@@ -90,8 +90,8 @@ exports.register = async (req: $Request, res: $Response) => {
 			tankName: 'Starter Tank',
 			userId: user.id,
 			components: ['moddable', null, null, 
-						null, null, null, 'fastTreads',
-						null, null, null]
+			null, null, null, 'fastTreads', 
+			null, null, null]
 		});
 
 		initTank.save((err: Error) => {
--- a/src/backend/controllers/userController.js
+++ b/src/backend/controllers/userController.js
@@ -79,7 +79,7 @@ exports.register = async (req: $Request, res: $Response) => {
 		let tank = new Tank();
 		tank.userId = user.id;
 		tank.tankName = 'Starter Tank';
-		tank.components = ['moddable', null, null, null, null, null, 'fastTreads', null, null, null];
+		tank.components = ['moddable', 'empty', 'empty', 'empty', 'empty', 'empty', 'fastTreads', 'empty', 'empty', 'empty'];
 		tank.casusCode = { 
 			boundingBox: { x: 0, y: 0, w: 64, h: 23 }, 
 			highlighted: false, 
--- a/src/backend/controllers/userController.js
+++ b/src/backend/controllers/userController.js
@@ -79,7 +79,7 @@ exports.register = async (req: $Request, res: $Response) => {
 		let tank = new Tank();
 		tank.userId = user.id;
 		tank.tankName = 'Starter Tank';
-		tank.components = ['moddable', 'empty', 'empty', 'empty', 'empty', 'empty', 'fastTreads', 'empty', 'empty', 'empty'];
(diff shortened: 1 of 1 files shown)


src/backend/controllers/userController.js
```diff
@@ -75,6 +76,33 @@ exports.register = async (req: $Request, res: $Response) => {
...
+		user = User.findOne({userName: userName, email: email}, (err: Error) => {
+			if (err) {
+				console.error(err.message);
+				return res
+					.status(500)
+					.json({ msg: 'Could not load created user' });
+			}
+		});
+
+		const initTank = new Tank({
```
SecondThread: Did you check opening up Casus for this? I think you might need to have a container block with an empty block inside as the default casus block for things to work correctly

SecondThread (commented):
Just one thing to double check, but if that works, then the code looks good to me

src/backend/controllers/userController.js
Derplime: Oh do I? Alright. I'll double check. Mind putting what the blank container will look like in a comment below me?

blaiadam: I think this is what he is referring to.

function getTestTank(id: number=1): Tank {
	const position=id===1?new Vec(20, -20):new Vec(50, 40);
	const toReturn: Tank = new Tank(
		position,
		new Chassis(id===1?'CHASSIS_5':'CHASSIS_2'), 
		new Treads(id===1?'TREAD_2':'TREAD_1'), 
		new Gun(id===1?'GUN_3':'GUN_4', false),
		new Gun(id===1?'GUN_9':'GUN_10', true),
		new Scanner(false, false, id===1?'MEDIUM':'SMALL'),
		new Jammer(id===1?'MEDIUM':'LARGE'),
		getEmptyCasusCode()
	);
	loadCasus(blocks => {toReturn.casusCode = blocks});
	return toReturn;
}

function getEmptyCasusCode(): CasusBlock {
	const childrenBlocks: Array<CasusBlock> = [];
	const container: ContainerBlock = new ContainerBlock(childrenBlocks);
	return container;
}

blaiadam: Lol it wouldn't print it in code view. Check TankLoader.js in master.

SecondThread: That might work? You would need to save the stringified version of that probably since this is after is is saved, but yeah, that's pretty much what it would look like. You can just use the default constructor (you don't need to pass in an empty array).

Derplime: I know what to do.  I'll just use the code from SittingDuck. Just have to export the tanks and copy the casusCode structure from the JSON file

Derplime:
Updated code with better logic and added base Casus code with empty bounding box. For what it's worth, opening Casus Editor was fine with both the empty bounding box and the null values for casusCode, but stuck with the bounding box since it's at least "something" that's populating the field.

src/backend/controllers/userController.js
```diff
@@ -66,12 +67,36 @@ exports.register = async (req: $Request, res: $Response) => {
...
+			return res
+				.status(500)
+				.json({ msg: 'Could not save the user to the DB.' });
+		}		
+
+		// Create initial tank for user
+		let tank = new Tank();
+		tank.userId = user.id;
+		tank.tankName = 'Starter Tank';
+		tank.components = ['moddable', null, null, null, null, null, 'fastTreads', null, null, null];
```
blaiadam: Change the null to 'empty'.

blaiadam (approved):
Thanks

PR #268, opened 2020-03-27
Added rendering of tank in armory

### LINKED ISSUE #190: Armory: Render selected tank in Armory (medium)
We currently aren't rendering the selected tank in the armory. We should be able to do this by creating a new ImageDrawer(src/battleground/ImageDrawer.js) but passing in a different canvas to draw on (here we would pass in a new canvas in the armory. We should still be able to use the same Tank class that we are using to battle with, we just wouldn't call Tank.executePhysics(), so it wouldn't like walk around the screen on us or anything.

If we wanted to, one thing we could do that would be really cool imo is have some custom Casus code that just animates the tank to turn the turret left, wait for a couple seconds, turn it back to the right, wait a couple seconds, et cetera. If we have this code saved, it would be really easy to put it into the tank and run it every 20th of a second, and then we would have a pretty idle animation for our tank when people are building their own tank.

SecondThread:
**Description**
Made tanks render in the armory. See #190 for details

Resolves #190

**Type of change**

Looks like this:
image

**Final Checklist:**


src/tanks/TankDisplay.js
```diff
@@ -0,0 +1,123 @@
...
+import Vec from '../casus/blocks/Vec.js';
+import {imageLoaderInit, addCallbackWhenImageLoaded} from '../battleground/ImageLoader.js';
+import './TankDisplay.css';
+
+type Props = {|
+	tankToDisplay: Tank
+|};
+
+type State = {|
+|};
```
JorgeVidal29: Does state have to be here?

src/tanks/TankDisplay.js
```diff
@@ -0,0 +1,123 @@
...
+		const canvas: HTMLCanvasElement = this.refs.canvas;
+		const targetWidth=canvas.clientWidth;
+		const targetHeight=targetWidth*8/16;
+		if (canvas.width !== targetWidth || canvas.height !== targetHeight) {
+			canvas.width = targetWidth;
+			canvas.height = targetHeight;
+		}
+	}
+
+	_lerp(a: number, b: number, time: number) {
```
JorgeVidal29: return type

src/tanks/TankDisplay.js
SecondThread: nope, thanks for catching

src/tanks/TankDisplay.js
SecondThread: Indeed, nice catch

JorgeVidal29 (approved):
Looks good to me

PR #272, opened 2020-03-28
Removed setState in constructors

blaiadam:
**Description**
Small change that moves any setState calls into componentDidMount. This is only recommended when the setState calls are in a fetch. Otherwise, do not put setState into componentDidMount.

More research needs to be done on this, but I don't think it is super important for right now. We will have to deal with the duplicate code.

**Type of change**

View code to see changes.

**Final Checklist:**


JorgeVidal29 (approved):
Ran flow and did some testing. Works fine

PR #304, opened 2020-03-29
Hardcode training bot ids in training arena

### LINKED ISSUE #290: Hardcode the training bot id's to the front end in the training arena
Currently the training bots have been created, but there is just placeholder text on the front end for what they actually are. This PR is to hardcode their IDs on the frontend so that when the user challenges them it starts a battle between the correct tank.

This also includes making changing the Battleground to be able to start a match given two tank ids, instead of a match id.

SecondThread:
**Description**
This PR hardcodes the training bot ids in the training arena so that we can later start battles against those training bots. Currently they aren't actually used for anything, but when we hook up training arena and battle arena to everything else, they will be.

Resolves #290

**Type of change**

Not really testable with the UI at the moment. This select works correctly:
image

And if you print the id of the tank that is selected when you click start, it matches what is in the DB.

**Final Checklist:**

### DIFF
--- a/src/trainingarena/TrainingArena.js
+++ b/src/trainingarena/TrainingArena.js
@@ -8,6 +8,26 @@ import { verifyLink } from '../globalComponents/verifyLink.js';
 import { verifyLogin } from '../globalComponents/verifyLogin.js';
 import setReturnToFromBattlegroundLink from '../battleground/setReturnToFromBattlegroundLink.js';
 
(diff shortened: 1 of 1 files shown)


blaiadam:
If you plan on using this for later, you should comment it out to remove the runtime error.
image

blaiadam (approved):
Looks great.

PR #361, opened 2020-04-04
Model revamp to use ObjectId types for related fields.

### LINKED ISSUE #358: Look into updating models properties that reference objectIds.
When most of the models were created initially, fields that referenced objectIds were typed as strings, and not as objectIds. This cleanup task will look into if changing the types to objectIds will affect the current implementation. If they do, the changes will be reverted, if not, the changes will be applied in a PR.

The significance behind this is the fact that the current implementations of those fields are resulting in conflicting results in data layer testing.

Derplime:
**Description**
All necessary models that have fields that reference objectIds are properly typed as ObjectIds in the model. All respective routes that accept these values have been edited to be validated for being MongoIds rather than strings.

The reason this was done is because it makes more sense to use this type in the models to enforce stricter validation in our API. This protects the routes that accept ObjectIds from random strings that are not properly structured. This also protects direct saves of invalid structured ObjectIds on documents (like internally through an API) due to the enforcement via the model. In extension this also irons out the data layer test implementation.

**A couple of things that need to be addressed due to this:**
* Test Data on Drive will be updated as of this PR.
* Your local documents have to be updated.
  * Any documents with the following non-null fields need to be updated to use ObjectIds:
    * User Model: favoriteTank
    * Tank Model: userId
    * Marketsale Model: sellerId
  * It's pretty easy to change them. Just go to the affected documents and change the type of the field to ObjectId. The field will adjust accordingly. Example picture is shown below:
image

**Resolves These Issues**
Resolves #358 

**Type of change**

- [x] This fix or feature will cause existing functionality to not work as expected. (Please elaborate in Description.)

* Just interact with the site as normal to see if any functions are messed up.
* I did my best to try everything that could be affected, and fortunately since we deal with the ObjectIds in the front-end/back-end integration, all investigated functions are working like they should.

**Final Checklist:**

- [x] My changes pass Flow, builds without errors and (if applicable) passes unit tests.

### DIFF
--- a/src/backend/controllers/marketController.js
+++ b/src/backend/controllers/marketController.js
@@ -67,8 +67,8 @@ exports.addMarketSale = async (req: Request, res: Response) => {
                 amount
             });
 
-            // Mark tank as being sold and save change to DB
-            tank.userId = 'Marketplace';
+            // Mark tank as being sold (give temporary ownership to master account) and save change to DB
+            tank.userId = MASTER_ID;
             await tank.save((err: Error) => {
                 if (err) {
                     console.error(err.message);
--- a/src/backend/controllers/tankController.js
+++ b/src/backend/controllers/tankController.js
@@ -44,6 +44,17 @@ exports.getFavorite = async (req: Request, res: Response) => {
 }
 
 exports.favoriteTank = async (req: Request, res: Response) => {
+	const errors = validationResult(req);
+
+	if (!errors.isEmpty()) {
+		// 400 is a bad request
+		console.error('Could not handle request');
+
+		return res
+			.status(400)
+			.json({ errors: errors.array() });
+	}
+
 	// the 'new' option means return the document after it has been updated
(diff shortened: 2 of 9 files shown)


src/backend/controllers/marketController.js
```diff
@@ -67,8 +67,8 @@ exports.addMarketSale = async (req: Request, res: Response) => {
                 amount
             });
 
-            // Mark tank as being sold and save change to DB
-            tank.userId = 'Marketplace';
+            // Mark tank as being sold (give temporary ownership to master account) and save change to DB
```
blaiadam, line 70: This is weird. Can't we just have the userId be from the user selling it?

blaiadam (approved):
LGTM

src/backend/controllers/marketController.js
Derplime: Post merge comment, the reason why we do that is because you shouldn't be able to access a tank that you're selling.

PR #383, opened 2020-04-05
Casus Recolor Built in Variables and Prompt previously used Variables

### LINKED ISSUE #372: Casus: Recolor Blocks for built-in names
Change the color of casus blocks so that variables that the user has created and variables that are system defaults should be different colors. This includes the following blocks:
- Get variable
- Set Variable

Also make the select variable popup show previous variables the user has created.

SecondThread:
**Description**
Constants are now a light-gray color. Built in variables are a sky-blue. User-created variables are the usual pink. Here is what things now look like:
image

Also, the list of suggested variables now includes all variables you have used before. Here is an example of what that looks like:
image

Passes flow:
image

**Resolves These Issues**
Resolves #372 
**Type of change**

**Final Checklist:**

### DIFF
--- a/src/armory/Armory.js
+++ b/src/armory/Armory.js
@@ -100,6 +100,10 @@ class Armory extends React.Component<Props, State> {
 	getTanks(): void {
 		getAllUsersTanks((successful, allTanks) => {
 			if (successful) {
+				if (allTanks.length === 0) {
+					console.log('Expected to have at least one tank!');
+					return;
+				}
 				// Always set the default selected tank to the newest tank.
 				const newSelectedTank = allTanks[allTanks.length-1];
 				setTankForCasus(newSelectedTank._id);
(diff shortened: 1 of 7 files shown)


src/armory/Armory.js
```diff
@@ -100,6 +100,10 @@ class Armory extends React.Component<Props, State> {
 	getTanks(): void {
 		getAllUsersTanks((successful, allTanks) => {
 			if (successful) {
+				if (allTanks.length === 0) {
+					console.log('Expected to have at least one tank!');
+					return;
+				}
```
Derplime, line 106: Did you just place this in since it was getting annoying entering the Armory if you didn't own a tank? That's the only thing that I can think of.

Derplime (approved):
Looks good. I loaded previously made Casus code and I was able to see the color differences and access previously made variables.

Just had one question below that just needs clarifying. Other than that it's good to go.

src/armory/Armory.js
SecondThread: Yeah, I was getting an error and it was confusing. I am pretty sure it was just because my database was screwed up though, it wasn't actually anything wrong with the code

PR #384, opened 2020-04-05
Better rounding and truncating

### LINKED ISSUE #81: Rounding a rediculously large double to an int might take an obscene amout of time
**Describe the bug**
If you have a double value that is rediculously big (order of magnitide of 1e100, for example) and then call round() on it, the interpriter will run forever. Well, techinically it will end eventually, but obviously this should run in O(1), instead of O(doubleValue/2^32) time.

**To Reproduce**
round a massive double to an int

SecondThread:
**Description**
Made rounding run in a reasonable amount of time. It turned out to be pretty easy, but now that we can run code it was easy enough to make sure that it worked as expected.

**Resolves These Issues**
Resolves #81 

**Type of change**

Same output as before:
image

Code that prints that:
image

Passes flow:
image

**Final Checklist:**

### DIFF
--- a/src/casus/interpreter/IntValue.js
+++ b/src/casus/interpreter/IntValue.js
@@ -97,7 +97,7 @@ class IntValue {
 		//validate that this number is in the range [-2^31, 2^31), and put it in there if it isn't
 		const minValue=-(2**31), maxValue=-minValue-1;
 		const modSize=2**32;
-		let x=input;
+		let x=input%modSize;
 
 		//these should each run at most once since the numbers stay small in intermediate calculations
 		while (x<minValue) {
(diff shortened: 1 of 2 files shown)


SecondThread:
This code also doesn't take forever to run:
image

src/casus/interpreter/IntValue.js
```diff
@@ -97,7 +97,7 @@ class IntValue {
 		//validate that this number is in the range [-2^31, 2^31), and put it in there if it isn't
 		const minValue=-(2**31), maxValue=-minValue-1;
 		const modSize=2**32;
-		let x=input;
+		let x=input%modSize;
```
blaiadam, line 100: lol, that's great

blaiadam (approved):
Passes flow and doesn't crash. Great job.

PR #386, opened 2020-04-05
Made Casus stop executing after 50,000 statements per frame

### LINKED ISSUE #184: Casus: Cut off execution of player's code after ~50000 instructions have run (medium?)
Currently there is no limit to the number of instructions you can run in a single casus frame. You can just crash the entire game if you (either accidentally or intentionally) have an infinite loop in your code. Obviously this isn't acceptable, so we need to count instructions and cut off the user's code in a deterministic way. There currently isn't a good way of doing this, which means we might have to touch the evaluation blocks of all ~65 blocks, which would be pretty annoying...

Overall, the amount of time this would take is a pretty big question mark because I haven't looked into how we are going to do this yet.

SecondThread:
**Description**
Made Casus code no longer crash the browser if it has an infinite loop. Here is a gif of a tank (bottom left) running with an infinite loop:
InfiniteLoop2

Passes flow:
image

**Resolves These Issues**
Resolves #184 

**Type of change**

Create a tank with an infinite (or insanely long) loop. When you battle with it, it shouldn't crash the browser.

**Final Checklist:**

### DIFF
--- a/src/casus/blocks/AndBlock.js
+++ b/src/casus/blocks/AndBlock.js
@@ -11,12 +11,12 @@ class AndBlock extends BinaryOperationBlock {
 	}
 
 	evaluate(): BooleanValue {
-		const lRes=verifyBoolean(this.lChild.evaluate());
+		const lRes=verifyBoolean(this.lChild.runEvaluate());
 		if (!lRes.val) {
 			return lRes;
 		}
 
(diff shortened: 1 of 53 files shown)


Derplime (approved):
Alrighty. I made a tank with an infinite loop and the cut off worked as it should. Nice job

PR #389, opened 2020-04-06
Added rightclick copying for Casus

### LINKED ISSUE #375: Casus: Change right-click to copy instead of select block
Currently right-clicking selects an contiguous range of blocks. This PR is to make left clicking do that, and to make right-clicking copy the block that is clicked.

SecondThread:
**Description**
Changed left-click to grab a contiguous block of casus blocks. It feels a little foreign to me now because I have been using the other system for a few weeks at this point, but if you guys don't like it, we can switch it back only selecting a single block by just changing a boolean, so that shouldn't be an issue.

You can also now copy an entire block and all its children by right-clicking on it. So that's pretty cool I think. It should massively speed up working with Casus I think. Here's kind of a sample of what that looks like:
image

Passes flow:
image

**Resolves These Issues**
Resolves #375 

**Type of change**

Go to the casus code for some tank.
Add some code that does whatever you feel like.
Right clicking on a block will now clone it.
Left clicking on a block will select that blocks and all blocks directly below it.

**Final Checklist:**

### DIFF
--- a/src/casus/CasusEditor.js
+++ b/src/casus/CasusEditor.js
@@ -12,6 +12,7 @@ import {isDefaultVariableName} from './userInteraction/defaultVariableNames.js';
 import './CasusEditor.css';
 import saveCasus from './saveCasus.js';
 import loadCasus from './loadCasus.js';
+import casusBlockDeepClone from './casusBlockDeepClone.js';
 
 type Props = {|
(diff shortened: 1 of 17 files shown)


blaiadam (approved):
Made a ton of for loops and print a rounded variable and no crash! Great job.

PR #391, opened 2020-04-06
Added Casus Functions

### LINKED ISSUE #371: Casus: Create Functions
Create Functions in Casus. This includes the following:
- Reorganizing some of the block bank so that there is a clean way to put functions.
- Create a Define Function block. The block should have a name.
- Include Define Function logic in the massive reviveCasusBlock switch statement
- Create an "Execute Function" block.
- Include Execute Function logic in the massive reviveCasusBlock switch statement
- Add logic to the chooseVariableName popup to choose the name of the function as well
- Add 'Defined functions' to the interpriter state, so that functions can be called.

SecondThread:
**Description**
Added everything required in this issue:
https://github.com/RickLeinecker/Fortuna/issues/371

Here is what it looks like:
image

- Recursion is supported. You can call a function from itself, or have a loop in function calls. This was a lot harder to do right than you would think because browsers have a really strict limit on how much your code can recurse, and the interpreter would need to evaluate this stuff recursively.
- You can minimize functions to easily compartmentalize your code.
- All variables are global. That is, there is not data stored on the stack, except for where you are in methods.

Here is what minimized methods look like:
image

Passes flow:
image

**Resolves These Issues**
Resolves #371 

**Type of change**

**Final Checklist:**

- [kinda] I have commented my code, particularly in hard-to-understand areas
- [kinda...] This change requires a update in the design document (if applicable).

### DIFF
--- a/src/casus/CasusEditor.js
+++ b/src/casus/CasusEditor.js
@@ -8,6 +8,8 @@ import Vec from './blocks/Vec.js';
 import SelectVariablePopup from './userInteraction/SelectVariablePopup.js';
 import GetVariableBlock from './blocks/GetVariableBlock.js';
 import SetVariableBlock from './blocks/SetVariableBlock.js';
+import DefineFunctionBlock from './blocks/DefineFunctionBlock.js';
+import CallFunctionBlock from './blocks/CallFunctionBlock.js';
 import {isDefaultVariableName} from './userInteraction/defaultVariableNames.js';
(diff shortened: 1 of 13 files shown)


blaiadam (approved):
This looks great. Make sure you check for conflicts if you merge your CasusCopying branch first. Also, I don't think users should be able to copy functions, but I'll leave it up to you.

PR #399, opened 2020-04-06
Fixed bug with blocks being destroyed on cancel

### LINKED ISSUE #388: Casus blocks are being removed when cancel is clicked when the aren't successfully placed
**Describe the bug**
When a casus block that needs to be renamed is dragged almost onto another block, but not released on a legal space, the block isn't placed (which is what is suppose to happen). However, if the user clicks "cancel" then a second block will get deleted

**To Reproduce**
1. drag a select variable block on top of an existing block.
2. click cancel when the dialog pops up.

**Expected behavior**
Either the dialog shouldn't open in the first place, or it should only delete the intended block.

**Additional context**
Caused by TODO on line 180 in CasusEditor. Remove TODO when this is fixed.

SecondThread:
**Description**
When you incorrectly place a block that needs to be renamed and then click cancel, there was previously a bug that caused whatever block you placed the block on to be destroyed. This PR fixes that. Here is a gif of it working as expected:
NotDeleteingOnCancel

Passes flow:
image

**Resolves These Issues**
Resolves #388 

**Type of change**

Try misplacing some block that would cause the popup to appear, and then click cancel. Everything should work as you would expect ti to

**Final Checklist:**

### DIFF
--- a/src/casus/CasusEditor.js
+++ b/src/casus/CasusEditor.js
@@ -190,7 +190,7 @@ class CasusEditor extends React.Component<Props, State> {
 	}
 
 	onCancelClicked() {
-		//delete variableblocktorename
+		//delete VariableBlockToRename
 		const block=this.state.variableBlockToRename;
 		if (block!=null) {
 			const toRemovePos=new Vec(block.boundingBox.x+1, block.boundingBox.y+1);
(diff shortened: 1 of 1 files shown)


blaiadam (approved):
Nice

PR #402, opened 2020-04-07
Added a much better debug system that doesn't crash the browser like console.log does

### LINKED ISSUE #373: Casus/Battleground: Make better debug logging that goes to screen instead of console
During training matches, make print statements print something to the screen instead of printing something to the console.

SecondThread:
**Description**
Previously to log debug statements, we were just printing them to the JS console. This was obviously not at all user friendly, and had the annoying side effect of crashing the user's browser if you logged way too many things (because the browser would stored all of the logged messages in memory). This is now fixed so that there is a (pretty nice) system that displays logged messages on the side of the screen. It looks like this:
image

Passes flow:
image

**Resolves These Issues**
Resolves #373 

**Type of change**

**Final Checklist:**

### DIFF
--- a/src/battleground/Battleground.css
+++ b/src/battleground/Battleground.css
@@ -1,4 +1,9 @@
 .battlegroundCanvas {
 	background-color: #fafafa;
-	width: 1300px; 
+	width: 1300px;
+}
+
+.battlegroundCanvasDiv {
+	width: 80%;	
+	display: inline-block;
 }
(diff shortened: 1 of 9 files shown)


SecondThread:
I guess I didn't mention this in the PR, but this only displays when the user is 1) in the training arena and 2) actually prints something. If either of those aren't true, it just stays hidden.

blaiadam (commented):
Isn't this in the healthbar PR?

blaiadam (approved):
Looks good.

PR #404, opened 2020-04-08
Finished a new, slick healthbar for 1v1s

### LINKED ISSUE #179: Game: Health bar (medium)
Create a health bar for tanks tanks on the battleground. This health bar should display both player's health, the names of both tanks, should indicate which items the players have left, and should indicate whether each player is jammed.

Implementation Details suggestions:
- Create a wrapper class for the Battleground (possibly BattlegroundContainer or something. An example is with CasusContainer). BattlegroundContainer can render both Battleground.js, and Healthbar.js
- Healthbar.js should have a canvas and be very similar in format to the Casus BlockBankBlockShelf as an example.
- In the rerender(), following the design of CasusContainer, you can get the canvasrenderingcontext2d of the canvas and draw stuff. With this you can draw images, geometric shapes, et cetera, so that should be plenty to make a solid health bar. You can also set fonts and draw text for the names (or if you want to label the amount of health each player has left, for example).
- In Healthbar's Props, pass in two Tank objects: the player's tank and the enemy's tank. Healthbar can look at these tanks to make decisions about rendering (i. e. how much health each player has left, whether they are jammed or not, et cetera)
- You can pass information (i. e. the two tanks) from Battleground to BattlegroundContainer to Healthbar with callbacks the same way CasusEditor passes to CasusContainer to BlockBankBlockShelf. You probably want the health bar to rerender every time Battleground calls the callback (i. e. after every update)

As far as how it looks goes, that's up to you, but here are a couple recommendations/things I think would be cool:
- You should be able to re-use art for mines/c4 that are loaded in the ImageLoader, so that shouldn't be an issue.
- I think it would be really cool if when a tank is jammed, you make a semi-transparent staticy box above his health bar with the text "JAMMED" or or something so the player knows what is going on.

SecondThread:
**Description**
Finished a new slick healthbar that shows the following things:
- All of the items the user has (mines, C4, powerups, missletracking beacons, et cetera)
- Whether the user is jammed
- The user's health and name
- A warning if the user's code has been cut off for running too many instructions
- A warning if the user's code has been cut off for too much recursion
Video of a match with the new healthbar:
https://drive.google.com/open?id=1pMidoklbQYyVMu45kdXXxeUxK85kq2sU

Passes flow:
image

**Resolves These Issues**
Resolves #179 

**Type of change**

**Final Checklist:**


blaiadam (approved):
Looks great. Thank you

PR #416, opened 2020-04-09
Prepare 3v3 match api

### LINKED ISSUE #365: update the prepareMatch api to support 3v3s
Receive the playerBeingChallengedUserId and the 3 favorited tanks of the challenger. Create a battleRecord holding the 3 tanks of each player.

N/A

**Additional context**
N/A

Airor4:
**Description**
Made an api to prepare the battleRecord for 3v3 battles. Returns the battleRecordId

**Resolves These Issues**
Resolves #365 

**Type of change**

- [X] New feature (adds functionality)

1) set a wager for the person who will be challenged
2) set the favoriteTanks to have 3 tanks using MongoDb compass
3) call the api with the personBeingChallengedId and 3 challengerTankIds
4) See if a battleRecord was created accordingly and you receive the battleRecordId

**Final Checklist:**

- [X] My changes pass Flow, build without errors and (if applicable) pass Jest tests.


src/models/battleRecordModel.js
```diff
@@ -40,60 +40,60 @@ const BattleRecord = new Mongoose.model('BattleRecord', new Mongoose.Schema({
     tankOne: {
         tankName: {
             type: String,
-            required: true
+            required: false
```
Derplime: Could just remove the required field entirely if we don't need it to be required. Same with the others.

Derplime (changes requested):
Works good on my end after setting everything up manually.
I'm going to be following your idea and just making a separate set of API for the 3v3 Favorites rather than adding the logic to the 1v1 ones that already work. You should probably do the same for reporting as well.
Just have that change I requested in the model then you're good.
EDIT: Adding picture of stored battle record with 3v3 teams
image

src/backend/controllers/battleController.js
```diff
@@ -128,6 +128,126 @@ exports.prepareMatch = async (req: Request, res: Response) => {
...
+		const tankTeamTwo = [];
+	
+		const personBeingChallenged = await User.findById(personBeingChallengedId);
+
+		if (personBeingChallenged == null) {
+			console.log('Could not find personBeingChallenged in DB');
+			return res
+				.status()
+				.json({ msg: 'Could not find user being challenged'});
+		} else if (personBeingChallenged.wager == 0) { // Check that they have a wager set
```
SecondThread: === please

src/backend/controllers/battleController.js
```diff
@@ -128,6 +128,126 @@ exports.prepareMatch = async (req: Request, res: Response) => {
...
+				return res
+					.status(404)
+					.json({ msg: 'Could not find one of the tanks of the challenger'});
+			}
+			tankTeamTwo.push(tank);
+		}
+
+		// Shouldnt need to check if null since we queried this earlier
+		// Also we need to do this so we can query the challenger user doc
+		const challengerId = await Tank.findById(challengerTankIds[0], 'userId');
```
SecondThread, line 198: you need to check to make sure that there are exactly 3 tanks in each list. If they give you a list of 0 tanks, it looks like this code would crash

src/backend/controllers/battleController.js
```diff
@@ -128,6 +128,126 @@ exports.prepareMatch = async (req: Request, res: Response) => {
...
+				.status(401)
+				.json({ msg: 'Challenger does not have enough money to wager'});
+		}
+
+		const newRecord = new BattleRecord({
+			userOne: personBeingChallengedId,
+			userTwo: challengerId.userId,
+			tankTeamOne: tankTeamOne,
+			tankTeamTwo: tankTeamTwo,
+			winner: -1,
```
SecondThread, line 222: This might be another PR, but sometime you have to assign an arena to a battlerecord as well.

src/models/battleRecordModel.js
```diff
@@ -40,60 +40,60 @@ const BattleRecord = new Mongoose.model('BattleRecord', new Mongoose.Schema({
     tankOne: {
```
SecondThread, line 45: talked about this offline, should probably just make an array of size 1.

SecondThread (changes requested):

src/backend/controllers/battleController.js
SecondThread: I guess if this is handled in the route, then this is a nonissue

Derplime:
@Derplime add logic to include maps in battle record:
image

Derplime:
@Derplime Also edit the battlerecord model to actually store Tank objects, since the 3v3 code is actually pushing the tanks to the array rather than making the defined objects from the model and pushing those.

src/backend/controllers/battleController.js
Derplime: Yeah it's tested in the route, I did a similar route check for the 3v3Favorites APIs and they worked for sizes 0, 1, and 2

Derplime: (worked as in they got flagged and stopped)

Derplime (approved):
Alright lgtm

src/backend/controllers/battleController.js
```diff
@@ -96,6 +97,13 @@ exports.prepareMatch = async (req: Request, res: Response) => {
 			eloExchanged: 0
 		});
 
+		const coinflip = Math.floor(Math.random() * 2);
+		if (coinflip === 0) {
```
SecondThread, line 101: You can clean this up a bit to just `if (Math.random() > 0.5)`

src/backend/controllers/battleController.js
```diff
@@ -128,6 +136,133 @@ exports.prepareMatch = async (req: Request, res: Response) => {
...
+		return res
+			.status(400)
+			.json({ errors: errors.array() });
+	}
+
+	const { personBeingChallengedId, challengerTankIds } = req.body;
+
+
+	try {
+		// Holds the array of each tank
```
SecondThread, line 154: array of each team, not of each tank

src/backend/controllers/battleController.js
```diff
@@ -128,6 +136,133 @@ exports.prepareMatch = async (req: Request, res: Response) => {
...
+			userOne: personBeingChallengedId,
+			userTwo: challengerId.userId,
+			tankTeamOne: tankTeamOne,
+			tankTeamTwo: tankTeamTwo,
+			winner: -1,
+			prizeMoney: (personBeingChallenged.wager * 2), // Each person puts in for the wager
+			eloExchanged: 0
+		});
+
+		const coinflip = Math.floor(Math.random() * 2);
```
SecondThread, line 227: same, you can clean this up a bit

SecondThread (approved):
Only minor suggestions. 

The on big thing that I didn't mention earlier is that if you are renaming the 1v1 battle, you want to make sure you change the API call to it on the front end to the corresponding name so that we aren't making illegal API calls in master when we challenge people.

PR #427, opened 2020-04-10
Toast messages for dailys

blaiadam:
**Description**
Whenever a user gets a daily stipend (through winning or setting their wager) a toast pops up and informs them. Also, in this branch is a fix for the wager not being taken away from user currency when wagering and a small replay table update.

**Resolves These Issues**

**Type of change**

1. Set a wager on a user that has not set a wager in the last 24 hours.
2. View toast.
3. Get a victory (ask David or I for code to have a tank win vs a sitting duck style tank).
4. View toast.

**Final Checklist:**

### DIFF
--- a/src/backend/controllers/userController.js
+++ b/src/backend/controllers/userController.js
@@ -522,7 +522,7 @@ exports.setWager = async (req: Request, res: Response) => {
 			else {
 				console.log('Wager updated and stipend added');
 				return res
-					.status(200)
+					.status(201)
 					.send(user);
 			}
 		}
--- a/src/backend/controllers/userController.js
+++ b/src/backend/controllers/userController.js
@@ -479,8 +479,9 @@ exports.setWager = async (req: Request, res: Response) => {
 
 		// If a user has never made a wager or its been 24hrs since last stipend give them their stipend and update wagerDate
 		const now = new Date();
+		let newBalance = 0;
 		if (user.wagerDate == null || (now - user.wagerDate) > aDay) {
-			const newBalance = user.money + 100;
+			newBalance = user.money + 100;
 			user.money = newBalance;
 			user.wagerDate = now;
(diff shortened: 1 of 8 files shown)


src/backend/controllers/userController.js
```diff
@@ -479,8 +479,9 @@ exports.setWager = async (req: Request, res: Response) => {
 
 		// If a user has never made a wager or its been 24hrs since last stipend give them their stipend and update wagerDate
 		const now = new Date();
+		let newBalance = 0;
```
SecondThread: If you are just using this as a boolean to check whether or not you applied the wager bonus, this is pretty confusing code. You should make it like:
`let appliedBonus=false;`
...
`appliedBonus=true;`

src/globalComponents/apiCalls/reportMatchResultAPICall.js
```diff
@@ -29,6 +29,9 @@ function reportMatchResultAPICall(winner: 0|1|2, matchId: string) {
 			}
 			else {
 				console.log('successfully logged result of match.');
+				if (getErrorFromObject(data) === 'First win of the day!') {
+					toast.success('First win of the day bonus added!');
```
SecondThread: it would be nice if you mentioned their current streak so that they get encouragement of it being multiple days

SecondThread (changes requested):

src/backend/controllers/userController.js
blaiadam: Good idea

src/globalComponents/apiCalls/reportMatchResultAPICall.js
blaiadam: Should be an easy setup

SecondThread (approved):
lgtm, thanks for fixing

PR #431, opened 2020-04-10
Fixed tank constructor to only start with the items it actually has

### LINKED ISSUE #403: Tanks are currently given all of the one-time-use items, instead of just the ones equipped
**Describe the bug**
Currently, all of the booleans in the gameState on whether a tank has each one-time item are just set to true in the constructor for debugging. They should be read from what the tank actually has.

**To Reproduce**
Not really applicable, but you can make a tank that used Nitro-recover for example even when it doesn't have it. You will see the particles appear, but really this should just be ignored, because the tank shouldn't have nitro-recover in the first place.

**Additional context**
This should just be a matter of adding extra fields to the constructor, and adding more logic to TankLoader.js

SecondThread:
**Description**
Previously, tanks started with all of the one-time items, reguardless of which ones were equipped. This PR fixes that so that now they only start with the ones they actually have.

Here is a screenshot of the fixed healthbar:
image

Passes flow:
image

**Resolves These Issues**
Resolves #403 

**Type of change**

Create a tank some subset of items. Exactly those items should show up in the healthbar and be usable.

**Final Checklist:**

### DIFF
--- a/src/tanks/Tank.js
+++ b/src/tanks/Tank.js
@@ -124,7 +124,6 @@ class Tank extends GameObject {
 		tankName: string,
 		_id: string,
 		userId: string,
-		
 	) {
 		super(position);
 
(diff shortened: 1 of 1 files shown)


Derplime (approved):
lgtm

PR #445, opened 2020-04-11
Duplicate items cannot be equipped (except mines)

### LINKED ISSUE #434: Armory: you can attach two C4s, nitrorepairs, overdrives, or missile tracking beacons to the same tank
**Describe the bug**
Currently there is nothing stopping you from attaching two c4s, nitro repairs, overdrives, or missile tracking beacons to the same tank. (You should be allowed to attach multiple mines though). Having multiple copies of these doesn't help at all, which makes it confusing that it is allowed. We probably shouldn't let users do this.

**To Reproduce**
1. Attach one c4 to your tank, and try to attach a second

**Expected behavior**
The c4 is grayed out since you already have attached it.

**Screenshots**
image

**Additional context**

blaiadam:
**Description**
There is a checkItems function that runs inside the checkPoints function when deciding whether or not to disable a component that can be equipped.
image

**Resolves These Issues**
Resolves #434 

**Type of change**

1. Go to Armory.
2. Make sure you have all items and your tank has very few points.
3. Equip an item besides the mine.
4. Go to another item slot and see that the option for the previously equipped component is now disabled.
5. Equip two mines to ensure it isn't disabled.

**Final Checklist:**

### DIFF
--- a/src/armory/Armory.js
+++ b/src/armory/Armory.js
@@ -162,6 +162,22 @@ class Armory extends React.Component<Props, State> {
 		});
 	}
 
+	// Checks the other items the tank has equipped in order to prevent two C4, nitro repair, etc.
+	checkItems(newComponent: TankComponent): boolean {
+		// Check if the new component is already in other slots, and that it is not a mine.
+		if (
+			((newComponent === this.state.selectedTank.parts[10].name) ||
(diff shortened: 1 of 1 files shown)


Derplime (approved):
Looks good. Nice job

PR #446, opened 2020-04-11
Unverified users get removed in 1 day and tank creation now in login.

### LINKED ISSUE #444: Make a way for unverified accounts to be removed from DB automatically.
There currently isn't a way for unverified users to be removed automatically from the DB if the account has been not verified after a certain amount of time. The only way to do this is manually.

Also there is a tight coupling issue with user creation such that when you make an unverified account, you make a tank too. So if the account is deleted, that unverified account's tank needs to be deleted as well.

So the solution I'm proposing here is:
- **Give UserModel a expirationDate field, similar to confirmationTokens, with an expiration of 24 hours.** So if the account isn't verified within 24 hours, the account is automatically removed from the database.
- **Remove the coupling with initial tank creation and user account creation.** 
  - This will be done by adding another field to UserModel to record last login, which will be set to `null` on default.
  - The initial tank creation will then be done on Login, removing the coupling with user creation, and it will check if the last login is `null`, it will make a new tank for the user, since it is the first time they've ever logged in successfully, then set the field to that time of login. If it hasn't been the first time, it just records that time of login. So there will only ever be one instance of this field being `null`, which is before the user logs in for the very first time.
- **Update User Email Confirmation to remove the expirationDate on the User.**
  - When a user confirms their account it feels pretty obvious at that point they want to keep it, so upon user confirmation, the API will also clear out the expirationDate field, so the user won't expire after 24 hours of its creation.

Fortunately, all these changes are backend models and APIs, and they don't affect any implementations on the front-end.

Derplime:
**Description**
This fixes the potential issue of many unverified accounts remaining in our DB until we remove them manually. Also fixes the coupling of initial tank creation with registering a user.

Changes include:
- UserModel
  - Fields for `expirationDate` and `lastLogin`
    - `lastLogin` set to `null` on default.
    - `expirationDate` set to `Date.now` as default, `expires` property set to 24 hours
- UserControllers
  - `registerUser` - Initial Tank creation removed from logic.
  - `login` - Initial Tank creation added to logic.
    - Initial tank is added if `lastLogin` is `null`, which only happens if the user is logging into their account **for the first time ever**. The `lastLogin` is then set to the time of login. Else, `lastLogin` is set to time of login.
  - `confirmToken` - Logic added to clear out the expiration date, preventing auto deletion.

**IMPORTANT NOTE: _When merged, this will cause an issue with existing accounts on your local databases. You can either clear your databases and use the backups on the Drive, or manually unverify the account using MongoDB Compass and reverify it using the APIs on Postman_**

**Resolves These Issues**
Resolves #444 

**Type of change**

1. Checkout branch
2. Temporarily edit files for easier testing 
  2a. `expires` property in the UserModel's `expirationDate` set to 30-60 seconds
  2b. In UserController.js, comment out token and email related logic in `registerUser` (Lines 80-133), and token related logic (don't forget in the DB query too) in `confirmToken` (Lines 272-280, the `_id: dbToken._userId` in the query, Lines 312-320)
3. `npm run build && npm run devserver`
In Postman:
4. Register a user
5. Confirm the user is created in MongoDB Compass with `null` lastLogin and a valid expirationDate
  5a. Confirm a tank isn't made for the user anymore at this point.
6. Wait 30-60 seconds (plus a minute), the time you set in UserModel.
7. Confirm the user gets auto deleted in Compass.
8. Repeat steps 4-5 with the same user that got deleted.
9. Confirm user email
10. Confirm that the expiration date got cleared in Compass.
11. Repeat step 6.
12. Confirm the user isn't deleted in Compass
13. Login as user
14. Confirm lastLogin is now set
15. Confirm a new tank is made for the user at this point.
16. Repeat steps 13-14
17. Confirm another tank isn't created on login.

**Final Checklist:**


Derplime:
_**IMPORTANT NOTE:**_ Test DB Backups and .env files are updated on Drive to account for this change. Be sure you backup anything that you might want to keep (likely tanks) before you import these.

Airor4 (approved):
Tested and it works. Might want to run it through David to make sure he approves the style but i'll approve for now.

PR #447, opened 2020-04-11
Show tank components being sold

### LINKED ISSUE #405: Marketplace tanks don't show the tank's components.
**Describe the Feature**
The user doesn't know what they are buying. Just a tank's name and its value.

**To Reproduce**
1. Go to marketplace.
2. Make a tank sale.
3. Log into other user.
4. Check tank sale.

**Expected behavior**
We need to display components of the tank.

**Screenshots**
N/A

**Additional context**
N/A

JorgeVidal29:
**Description**
Tanks are now shown in the marketplace

**Resolves These Issues**
Resolves #405 

**Type of change**

Made sure to have a tank in the marketplace
Went to marketplace
Saw the tanks that are being sold displayed

**Final Checklist:**

### DIFF
--- a/src/globalComponents/apiCalls/marketPlaceAPIConnections.js
+++ b/src/globalComponents/apiCalls/marketPlaceAPIConnections.js
@@ -101,7 +101,8 @@ function getMarketTanks(userId: string, onLoad:(tanks: Array<SaleObject>) => voi
 						sale.salePrice,
 						sale.amount,
 						sale.sellerId,
-						sale._id
+						sale._id,
+						sale.itemId._id
 					));
 				} 
 				onLoad(itemsForSale);
--- a/src/globalComponents/apiCalls/marketPlaceAPIConnections.js
+++ b/src/globalComponents/apiCalls/marketPlaceAPIConnections.js
@@ -69,7 +69,7 @@ function getMarketSales(userId: string, onLoad:(sales: Array<SaleObject>) => voi
 						sale.amount,
 						sale.sellerId,
 						sale._id,
-						"0"
+						null
 					));
 				}
 				onLoad(itemsForSale);
--- a/src/globalComponents/apiCalls/marketPlaceAPIConnections.js
+++ b/src/globalComponents/apiCalls/marketPlaceAPIConnections.js
@@ -68,7 +68,8 @@ function getMarketSales(userId: string, onLoad:(sales: Array<SaleObject>) => voi
 						sale.salePrice,
 						sale.amount,
 						sale.sellerId,
-						sale._id
+						sale._id,
+						"0"
 					));
 				}
 				onLoad(itemsForSale);
--- a/src/globalComponents/apiCalls/tankAPIIntegration.js
+++ b/src/globalComponents/apiCalls/tankAPIIntegration.js
@@ -228,7 +228,7 @@ function getTankById(tankId: string, onLoad:(tank: Tank) => void): void {
 					data.components,
 					data.casusCode,
 					data.isBot,
-					data.userid,
+					data.userId,
 					data.tankName
 				);
 				onLoad(getTank(tank));
--- a/src/globalComponents/apiCalls/tankAPIIntegration.js
+++ b/src/globalComponents/apiCalls/tankAPIIntegration.js
@@ -203,6 +203,40 @@ function deleteTank(tankId: string, onLoad:() => void): void {
 	);
 }
 
+//Input: Id of a tank
+//Output: a tank object of the tank whose id equals the id that was passed in
+function getTankById(tankId: string, onLoad:(tank: Tank) => void): void {
+	const responsePromise: Promise<Response> = fetch('/api/tank/getTankById/' + tankId, {
+		method: 'GET',
+		headers: {
+			'Access-Control-Allow-Origin': '*',
+			'Content-Type': 'application/json',
+			'Access-Control-Allow-Credentials': 'true',
+			'x-auth-token': getLoginToken()
+		},
+	});
+	responsePromise.then(
+		response => response.json().then(data => {
+			if (response.status !== 200) {
+				console.log(response.status);
+				console.log(data);
+				toast.error(getErrorFromObject(data));
+			}
+			else {
+				const tank = new BackendTank(
+					data._id,
+					data.components,
+					data.casusCode,
+					data.isBot,
+					data.userid,
+					data.tankName
+				);
+				onLoad(getTank(tank));
+			}
+		})
+	)
+}
+
 export {
 	getFavoriteTank,
 	setFavoriteTankId,
@@ -210,6 +244,7 @@ export {
 	updateTank,
 	getAllUsersTanks,
 	createTank,
-	deleteTank
+	deleteTank,
+	getTankById
 }
 
--- a/src/globalComponents/apiCalls/tankAPIIntegration.js
+++ b/src/globalComponents/apiCalls/tankAPIIntegration.js
@@ -205,11 +205,7 @@ function deleteTank(tankId: string, onLoad:() => void): void {
 
 //Input: Id of a tank
 //Output: a tank object of the tank whose id equals the id that was passed in
-function getTankById(tankId: ?string, onLoad:(tank: Tank) => void): void {
-	if(tankId == null) {
-		console.log("Tank Id is null");
-		return;
-	}
+function getTankById(tankId: string, onLoad:(tank: Tank) => void): void {
 	const responsePromise: Promise<Response> = fetch('/api/tank/getTankById/' + tankId, {
 		method: 'GET',
 		headers: {
--- a/src/globalComponents/apiCalls/tankAPIIntegration.js
+++ b/src/globalComponents/apiCalls/tankAPIIntegration.js
@@ -205,7 +205,11 @@ function deleteTank(tankId: string, onLoad:() => void): void {
 
 //Input: Id of a tank
 //Output: a tank object of the tank whose id equals the id that was passed in
-function getTankById(tankId: string, onLoad:(tank: Tank) => void): void {
+function getTankById(tankId: ?string, onLoad:(tank: Tank) => void): void {
+	if(tankId == null) {
+		console.log("Tank Id is null");
+		return;
+	}
 	const responsePromise: Promise<Response> = fetch('/api/tank/getTankById/' + tankId, {
 		method: 'GET',
 		headers: {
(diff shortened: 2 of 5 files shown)


src/marketplace/ListingsView.js
```diff
@@ -78,6 +83,10 @@ class ListingsView extends React.Component<Props, State> {
 	getMarketSalesForTanks(): void  {
 		getMarketTanks(this.state.userId, sales => {
 				this.setState({itemsForSale: sales.filter(sale => !(allComponents.includes(sale.name)))}); 
+				//Once we have all of the tanks being sold add them to the array that is going to house those tanks
+				for(let i = 0; i < sales.length; i++) {
```
blaiadam: When you have a lot of tanks in the marketplace, you will be updating the state a ton. Change this to create an array of Tanks first, and then set the state.

blaiadam (changes requested):
Looks good. I have one change requested, fix that and I'll approve.

src/globalComponents/apiCalls/marketPlaceAPIConnections.js
```diff
@@ -68,7 +68,8 @@ function getMarketSales(userId: string, onLoad:(sales: Array<SaleObject>) => voi
 						sale.salePrice,
 						sale.amount,
 						sale.sellerId,
-						sale._id
+						sale._id,
+						"0"
```
SecondThread: In JS single quotes are preferred according to our style guide.

Also, this should probably be an empty string

src/globalComponents/typesAndClasses/SaleObject.js
```diff
@@ -6,13 +6,15 @@ class SaleObject {
 	amount: number;
 	sellerId: string;
 	saleId: string;
+	tankId: string;
```
SecondThread: If sale objects are also used for item sales, then this should definitely be nullable. Otherwise you have a very confusing sentinel value...

src/globalComponents/apiCalls/marketPlaceAPIConnections.js
SecondThread: On second thought, it should definitely be null, not empty string. I didn't fully read what this was being used for.

src/marketplace/ListingsView.js
```diff
@@ -77,7 +82,17 @@ class ListingsView extends React.Component<Props, State> {
...
+				//Once we have all of the tanks being sold add them to the array that is going to house those tanks
+				const tanksForSale: Array<Tank> = [];
+				for(const tank of sales) {
+					getTankById(tank.tankId, tankReturned => {
+						tanksForSale.push(tankReturned);
+					});
+				}
+				this.setState({
+					itemsForSale: sales.filter(sale => !(allComponents.includes(sale.name))),
+					tanksBeingShown: tanksForSale
```
SecondThread: I talked to Adam about this, but this is very confusing and probably doesn't do what you want it to in some cases. This array isn't actually full yet, so you are about to call render without the stuff you need in this array, and then it will get put into the array after the fact

SecondThread: The solution is to not call setstate until after the API call you are waiting for returns

SecondThread (changes requested):

src/globalComponents/apiCalls/tankAPIIntegration.js
```diff
@@ -203,13 +203,52 @@ function deleteTank(tankId: string, onLoad:() => void): void {
 	);
 }
 
+//Input: Id of a tank
+//Output: a tank object of the tank whose id equals the id that was passed in
+function getTankById(tankId: ?string, onLoad:(tank: Tank) => void): void {
+	if(tankId == null) {
```
SecondThread: If this method doesn't allow tankId to be null, tankId shouldn't be nullable, and you don't need this if statement. Wherever you call this method, you need to be careful that you don't pass null into it.

src/marketplace/ShowTankPopup.js
```diff
@@ -0,0 +1,74 @@
...
+	tankBeingShown: Tank
+|}
+
+// This popup takes an id of a tank, gets that tank, and shows it
+class ShowTankPopup extends React.Component<Props, State> {
+
+	constructor(props: Props) {
+		super(props);
+		//This blank tank is a placeholder for the tank that we are goiing to pull from the show tank function
+		const blankTank: BackendTank = new BackendTank(
```
SecondThread: This is very sketchy. Instead of using sentinel values, you should make tankBeingShown nullable.

blaiadam (approved):
Looks good

src/marketplace/ListingsView.js
```diff
@@ -86,21 +87,38 @@ class ListingsView extends React.Component<Props, State> {
...
+			// Handle tank and components different to display tank 
+			// Have to make sure that the tanks are being ready to be shown
+			if(this.props.sellerType === 'tanks') {
+				cards.push(
+					<div className="card mb-2" key={i}>
+						<div className="card-body">
+							<h5 className="card-title">Item to buy: {toTitleCase(this.state.itemsForSale[i].name)}</h5>
+							<h5 className="card-title">Price: ${this.state.itemsForSale[i].price}</h5>
+							<h5 className="card-title">Quantity: {this.state.itemsForSale[i].amount}</h5>
+							<ShowTankPopup tankIdToShow={this.state.itemsForSale[i].tankId}></ShowTankPopup>
```
SecondThread: You can do something like:
```
const tankIdForItem = this.state.itemsForSale[i].tankId;
if (tankIdForItem == null) {
  throw new Error('something describing why this is illegal');
}
```
above the JSX.
Then you don't need to make ShowTankPopup.props.tankIdToShow nullable. (It doesn't make sense to me at least for it to be nullable because it doesn't make sense to show a tank with null as its id)

src/marketplace/ListingsView.js
```diff
@@ -86,21 +87,38 @@ class ListingsView extends React.Component<Props, State> {
...
+							<button className="btn btn-success mt-2" onClick={() => this.buyItem(this.state.itemsForSale[i].sellerId, this.state.itemsForSale[i].saleId)}>Buy</button>
+						</div>
 					</div>
-				</div>
-			)
+				)
+			}
+			else if(this.props.sellerType !== 'tanks') {
```
SecondThread: uh, isn't that what 'else' means?

if (x) 
  ...
else if (!x)
  ...
??

src/marketplace/ShowTankPopup.js
```diff
@@ -0,0 +1,74 @@
...
+			<div>
+				<button className="btn mt-2" onClick={() => this.showTank()}>
+					See Tank
+				</button>
+				<Popup 
+					open={this.state.showingTank}
+					onClose={() => this.handleCloseClick()}
+				>
+					<div className="popup">
+						<TankDisplay tankToDisplay={this.state.tankBeingShown} smallTank={false} />
```
SecondThread: If, after you make the above changes, you are about to get a flow error saying that this.state.tankBeingShown can be null, but tankToDisplay can't be, then you can check if it is null, and if it is show like the text "loading tank..." instead for instance

SecondThread (changes requested):
Pretty small stuff, but I would like if you could fix it before pushing

src/globalComponents/apiCalls/tankAPIIntegration.js
JorgeVidal29: If we don't do that flow gives that error we talked about

JorgeVidal29: Fixing the below fixed this

src/marketplace/ListingsView.js
```diff
@@ -86,21 +87,44 @@ class ListingsView extends React.Component<Props, State> {
...
-						<h5 className="card-title">Item to buy: {toTitleCase(this.state.itemsForSale[i].name)}</h5>
-						<h5 className="card-title">Price: ${this.state.itemsForSale[i].price}</h5>
-						<h5 className="card-title">Quantity: {this.state.itemsForSale[i].amount}</h5>
-						<button className="btn btn-success mt-2" onClick={() => this.buyItem(this.state.itemsForSale[i].sellerId, this.state.itemsForSale[i].saleId)}>Buy</button>
+			// Handle tank and components different to display tank 
+			// Have to make sure that the tanks are being ready to be shown
+			if(this.props.sellerType === 'tanks') {
+				const tankId = this.state.itemsForSale[i].tankId;
+				if(tankId == null) {
+					console.log("Tank Id is null");
```
SecondThread: You should throw an error here so that we notice

src/marketplace/ListingsView.js
```diff
@@ -86,21 +87,44 @@ class ListingsView extends React.Component<Props, State> {
...
+				}
+				else {
```
SecondThread: Dont need the else if we are crashing in the if statement

src/marketplace/ListingsView.js
```diff
@@ -86,21 +87,44 @@ class ListingsView extends React.Component<Props, State> {
...
+						<div className="card-body">
+							<h5 className="card-title">Item to buy: {toTitleCase(this.state.itemsForSale[i].name)}</h5>
+							<h5 className="card-title">Price: ${this.state.itemsForSale[i].price}</h5>
+							<h5 className="card-title">Quantity: {this.state.itemsForSale[i].amount}</h5>
+							<button className="btn btn-success mt-2" onClick={() => this.buyItem(this.state.itemsForSale[i].sellerId, this.state.itemsForSale[i].saleId)}>Buy</button>
+						</div>
 					</div>
-				</div>
-			)
+				)
```
SecondThread: semicolon please

src/marketplace/ListingsView.js
```diff
@@ -86,21 +87,44 @@ class ListingsView extends React.Component<Props, State> {
...
+						<div className="card mb-2" key={i}>
+							<div className="card-body">
+								<h5 className="card-title">Tank to buy: {toTitleCase(this.state.itemsForSale[i].name)}</h5>
+								<h5 className="card-title">Price: ${this.state.itemsForSale[i].price}</h5>
+								<h5 className="card-title">Quantity: {this.state.itemsForSale[i].amount}</h5>
+								<ShowTankPopup tankIdToShow={tankId}></ShowTankPopup>
+								<button className="btn btn-success mt-2" onClick={() => this.buyItem(this.state.itemsForSale[i].sellerId, this.state.itemsForSale[i].saleId)}>Buy</button>
+							</div>
+						</div>
+					)
```
SecondThread: semicolon please

SecondThread (approved):
Please fix suggestions before pushing

PR #448, opened 2020-04-11
Replay table now shows more information

### LINKED ISSUE #442: Display whether a battle was 1v1 or 3v3 on the Battle Arena Table
Currently we aren't displaying whether a match was a 1v1 match or a 3v3 match in the battle arena table, but we should be.

blaiadam:
**Description**
I updated the replay class to reflect the battle record model better for 3v3's. With this change, The replay table now can tell the difference between 1v1 and 3v3 battles.

Some additional changes, the prize money and elo exchanged now shows a + or - depending on if you won or lost. Also, the result no longer says "user won", instead it just says "Win" or "Loss".

image

**Resolves These Issues**
Resolves #442 

**Type of change**

1. Got to battle arena.
2. Make sure you have both a win and a loss.
3. View replay table updates and make sure they are concurrent with the changes from the battle (prize for victories shows + not -).

**Final Checklist:**

### DIFF
--- a/src/battlearena/Replays.js
+++ b/src/battlearena/Replays.js
@@ -58,8 +58,9 @@ class Replays extends React.Component<Props, State> {
 	render(): React.Node {
 		return (
 			<div className="replayList">
-				<h4>{this.state.myUsername}'s Battle Record</h4>
+				
 				<div className="replayTable">
+					<h4>{this.state.myUsername}'s Battle Record</h4>
 					<table>
 						<thead>
 							<tr>
--- a/src/battlearena/Replays.js
+++ b/src/battlearena/Replays.js
@@ -58,7 +58,6 @@ class Replays extends React.Component<Props, State> {
 	render(): React.Node {
 		return (
 			<div className="replayList">
-				
 				<div className="replayTable">
 					<h4>{this.state.myUsername}'s Battle Record</h4>
 					<table>
--- a/src/battlearena/Replays.js
+++ b/src/battlearena/Replays.js
@@ -31,12 +31,12 @@ class Replays extends React.Component<Props, State> {
 		});
 	}
 
-	getWinner(replay: Replay): string {
+	getMatchResult(replay: Replay): string {
 		switch(replay.winner) {
 			case 2:
-				return replay.playerTwoName + ' won';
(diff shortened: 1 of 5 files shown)


src/battlearena/Replays.js
```diff
@@ -57,32 +57,32 @@ class Replays extends React.Component<Props, State> {
...
+							<th>Opponent</th>
+							<th>Result</th>
+							<th>Replay</th>
+							<th>Prize</th>
+							<th>Elo</th>
+							<th>Type</th>
+						</tr>
+					</thead>
+					<tbody>
+						{this.state.replays.slice(0).reverse().map(replay =>
```
SecondThread, line 74: you don't need to change it, just so I understand what is going on here: the slice(0) is so that reverse doesn't flip the array in your state, right?

src/globalComponents/typesAndClasses/Map.js
```diff
@@ -0,0 +1,5 @@
+//@flow strict
```
SecondThread: This is a duplicate of the ArenaType class I think

src/globalComponents/typesAndClasses/Replay.js
```diff
@@ -1,24 +1,31 @@
...
-		tankTwoName: string,
+		tankOneName: ?string,
+		tankTwoName: ?string,
 		playerOneName: string,
 		playerTwoName: string,
 		winner: -1|0|1|2,
 		prizeMoney: number,
 		eloExchanged: number,
 		replayId: string,
+		map: Map,
```
SecondThread: Yeah, we should probably be consistant with this and either use Map everywhere or use ArenaType everywhere.

SecondThread (approved):
Mostly looks good, the one thing I would change is that the map class is a duplicate of the ArenaType class

src/globalComponents/typesAndClasses/Map.js
blaiadam: Ah, I couldn't find one. I'll update it.

src/battlearena/Replays.js
JorgeVidal29: I think a comment may be useful here

PR #481, opened 2020-04-14
Design updates

blaiadam:
**Description**
Check out the changes and see if they look good.

**Resolves These Issues**
No issue to resolve, just updating according to feedback.

**Type of change**

1. View changes in code and view suggestions in my second comment in the Feeback issue.
2. Check website to see new changes.

**Final Checklist:**

### DIFF
--- a/src/backend/controllers/userController.js
+++ b/src/backend/controllers/userController.js
@@ -160,12 +160,6 @@ exports.login = async (req: Request, res: Response) => {
 	try{
 		// See if User exists - might change this to const
 		const user = await User.findOne({ userName: userName });
-		if(user == null){
-			return res
-				.status(401)
-				.json({ msg: 'The user name ' + userName + 
-					' is not associated with any account.' });
-		}
 
 		// Checks if the plaintext password matches the hashed pass form db
 		const isMatch = await bcrypt.compare(password, user.password);
--- a/src/backend/controllers/userController.js
+++ b/src/backend/controllers/userController.js
@@ -171,10 +171,10 @@ exports.login = async (req: Request, res: Response) => {
 		const isMatch = await bcrypt.compare(password, user.password);
 
 		// If the credentials don't match it will return a boolean false
-		if (!isMatch) {
+		if (!isMatch || user == null) {
 			return res
 				.status(401)
-				.json({ msg: 'Invalid Password. Please try again.' });
+				.json({ msg: 'Incorrect username or password' });
 		}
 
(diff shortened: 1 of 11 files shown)


src/backend/controllers/userController.js
```diff
@@ -164,7 +164,7 @@ exports.login = async (req: Request, res: Response) => {
 			return res
 				.status(401)
 				.json({ msg: 'The user name ' + userName + 
-					' is not associated with any account. Double-check your user name and try again.' });
+					' is not associated with any account.' });
```
Derplime: This change is fine, but now I'm just generally questioning what messages we give as responses and whether in this case we should give a more generic one (security me musing to myself)

src/backend/controllers/userController.js
```diff
@@ -182,8 +182,7 @@ exports.login = async (req: Request, res: Response) => {
 			return res
 				.status(401)
 				.json({ type: 'email-not-verified', msg: 'Your account has not been verified. ' + 
-					'Check your email for your verification token. ' +
-					'If your token has expired, request a new one.' });
+					'Check your email for verification, or request a new one.' });
```
Derplime: 'Check your email **to verify your account**,...' would sound nicer but it's more a suggestion.

src/backend/controllers/userController.js
```diff
@@ -277,9 +276,9 @@ exports.confirmToken = async (req: Request, res: Response) => {
 			return res
 				.status(400)
 				.json({ type: 'not-verified',
-					msg: 'The token you are using is not a valid token. ' +
+					msg: 'The token you are using is not valid. ' +
 					'Check your confirmation email and try again. ' + 
-					'Otherwise, your token may have expired.' });
+					'Otherwise, request a new confirmation email at the Login page.' });
```
Derplime: I think it would be better to remove the "Otherwise," here and just say "Request a new....."

Derplime (changes requested):
Added a couple of musings and suggestions.

src/backend/controllers/userController.js
blaiadam: I like this too.

src/backend/controllers/userController.js
Derplime: Not that this should be changed now, since it's fine as it is, but looking back at it now for this check we should have grouped the check for a user and pw match as a single condition and send a single response back if either failed rather than check them in order.

Derplime (approved):
Looks good

PR #498, opened 2020-04-15
Fixing saving last tank bug

### LINKED ISSUE #478: Heading from training arena back to casus sometimes switches tanks
**Describe the bug**
It is rare, but sometimes when I am heading back to casus from the training arena, I get a different tank.

**To Reproduce**
1. Go to training arena through casus.
2. Go back to casus.

**Expected behavior**
It should retain the cookie, I haven't changed my selected tank in the training arena

**Screenshots**
N/A

**Additional context**
I need to test this more to effectively reproduce.

SecondThread:
**Description**
Fixed the bug with the casus tank getting reset after a match. There was also another bug where the tank you set was not getting saved correctly in Armory. This was because we were using sentinal values and passing around non-null tanks, so the cookie was getting cleared. I fixed that here, so now thing should work correctly.

image

**Resolves These Issues**
Resolves #478 

**Type of change**

Watch a match to finish in the training arena, and then immediately click "back to casus". You should consistently be on the correct tank now.

**Final Checklist:**

### DIFF
--- a/src/battleground/Battleground.js
+++ b/src/battleground/Battleground.js
@@ -55,7 +55,7 @@ const arenaWidth: {[ArenaType]: number} = {
 }
 
 const matchLengthForArena: {[ArenaType]: number} = {
-	DIRT: INTRO_LENGTH + 30*60,
+	DIRT: INTRO_LENGTH + 30*1,
 	HEX: INTRO_LENGTH + 30*60,
 	LUNAR: INTRO_LENGTH + 30*100,
 	CANDEN: INTRO_LENGTH+ 30*100,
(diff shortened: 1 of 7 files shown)


src/battleground/Battleground.js
```diff
@@ -55,7 +55,7 @@ const arenaWidth: {[ArenaType]: number} = {
 }
 
 const matchLengthForArena: {[ArenaType]: number} = {
-	DIRT: INTRO_LENGTH + 30*60,
+	DIRT: INTRO_LENGTH + 30*1,
```
Derplime: Is this supposed to be different from the rest?

Derplime, line 58: Is this supposed to be different from the rest?

Derplime (approved):
I have one comment included otherwise it's good.

src/battleground/Battleground.js
SecondThread: oops, no I just had that in there for testing so I didn't have to wait for the whole minute. Nice catch.

PR #505, opened 2020-04-15
Fixed username/password checking

### LINKED ISSUE #501: Server error on incorrect credentials
**Describe the bug**
The message for incorrect credentials is broken

**To Reproduce**
1. Put incorrect login credentials.
2. View error

**Expected behavior**
Should print "Incorrect username or password, please try again"

**Screenshots**
image

**Additional context**

Derplime:
**Description**
When we combined the logic to check for a valid username and password match, we forgot that the bcrypt check used `user.password` as an argument, so if the user const from the query was `null`, it broke the bcrypt call, causing the 500 server error status. This PR handles the `null` user query result in between the query and the bcrypt call.

**Resolves These Issues**
Resolves #501 

**Type of change**

1. Attempt to login with your user, but:
  1a. Have the right password
  1b. Misspell your username.
2. Confirm the Server Error response doesn't come back, and the Incorrect Username or PW response comes instead,

**Final Checklist:**

### DIFF
--- a/src/backend/controllers/userController.js
+++ b/src/backend/controllers/userController.js
@@ -162,7 +162,7 @@ exports.login = async (req: Request, res: Response) => {
 		const user = await User.findOne({ userName: userName });
 
 		// Internal password constant
-		const passCheck = (user) ? user.password : "";
+		const passCheck = user?.password ?? '';
 
 		// Checks if the plaintext password matches the hashed pass form db
 		const isMatch = await bcrypt.compare(password, passCheck);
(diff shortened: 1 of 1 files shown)


src/backend/controllers/userController.js
```diff
@@ -161,8 +161,11 @@ exports.login = async (req: Request, res: Response) => {
 		// See if User exists - might change this to const
 		const user = await User.findOne({ userName: userName });
 
+		// Internal password constant
+		const passCheck = (user) ? user.password : "";
```
SecondThread: sketchy null check. `user!=null` preferred if you are trying to do this. Also single quotes are preferred in JS unless the string contains a single quote.

Even better is:
`const passCheck = user?.password ?? '';`

SecondThread (changes requested):

SecondThread (approved):
lgtm

PR #506, opened 2020-04-16
Added changing the casus code from one tank to another

### LINKED ISSUE #494: Copy casus code from other tank button in armory
Copy casus code from one tank to another in armory

**Additional context**

JorgeVidal29:
**Description**
Players can now copy the casus code from one tank to another

**Resolves These Issues**
Resolves #494 

**Type of change**

Went to Armory
Clicked on change casus code
Selected a different tank
Clicked on copy
Went to the casus code to see the difference

**Final Checklist:**

### DIFF
--- a/src/armory/Armory.js
+++ b/src/armory/Armory.js
@@ -12,6 +12,7 @@ import DeleteTankPopup from './DeleteTankPopup.js';
 import SelectTank from '../globalComponents/SelectTank.js';
 import SetWagerPopup from './SetWagerPopup.js';
 import RenameTankPopup from './RenameTankPopup.js';
+import CopyCasusCodePopup from './CopyCasusCodePopup';
 import { ToastContainer, toast } from 'react-toastify';
 // Functions
 import { getComponentPoints, getComponentType } from '../globalComponents/GetInventoryInfo.js';
@@ -346,7 +347,13 @@ class Armory extends React.Component<Props, State> {
 							<DeleteTankPopup
 								tank={selectedTank}
 							/>
-						}
+						}&emsp;
+						{selectedTank==null?<div></div>:
+							<CopyCasusCodePopup
+								selectedTank={selectedTank}
+								usersTanks={this.state.allTanks}
+							/>
+						}&emsp;
 					</div>
 					<br/><br/>
 					<SetWagerPopup
--- a/src/armory/Armory.js
+++ b/src/armory/Armory.js
@@ -332,10 +332,10 @@ class Armory extends React.Component<Props, State> {
 					</Link>
 					<br/><br/>
 					{selectedTank==null?<div></div>:
-							<CopyCasusCodePopup
-								selectedTank={selectedTank}
-								usersTanks={this.state.allTanks}
-							/>
+						<CopyCasusCodePopup
+							selectedTank={selectedTank}
+							usersTanks={this.state.allTanks}
+						/>
 					}
 					<br/><br/><br/>
 					<h5>Tank Options</h5>
--- a/src/armory/Armory.js
+++ b/src/armory/Armory.js
@@ -325,10 +325,18 @@ class Armory extends React.Component<Props, State> {
 						/>
 					}
 					<br/><br/>
+					<label>Edit Code</label>
+					<br/>
 					<Link to={verifyLink("/Casus")}>
 						<button className="primarybtn">Casus</button>
 					</Link>
-					<label>&emsp;Edit Code</label>
+					<br/><br/>
+					{selectedTank==null?<div></div>:
+							<CopyCasusCodePopup
+								selectedTank={selectedTank}
+								usersTanks={this.state.allTanks}
+							/>
+					}
 					<br/><br/><br/>
 					<h5>Tank Options</h5>
 					<div className="row rowPadding">
@@ -347,13 +355,7 @@ class Armory extends React.Component<Props, State> {
 							<DeleteTankPopup
 								tank={selectedTank}
 							/>
-						}&emsp;
-						{selectedTank==null?<div></div>:
-							<CopyCasusCodePopup
-								selectedTank={selectedTank}
-								usersTanks={this.state.allTanks}
-							/>
-						}&emsp;
+						}
 					</div>
 					<br/><br/>
 					<SetWagerPopup
--- a/src/armory/CopyCasusCodePopup.js
+++ b/src/armory/CopyCasusCodePopup.js
@@ -59,6 +59,16 @@ class CopyCasusCodePopup extends React.Component<Props, State> {
 		}
 	}
 
+	removeSelectedTankFromList(): Array<Tank> {
+		const listOfTanksWithoutCurrentTank = [];
+		for(let i = 0; i < this.props.usersTanks.length; i++) {
+			if(this.props.usersTanks[i]._id !== this.props.selectedTank._id) {
+				listOfTanksWithoutCurrentTank.push(this.props.usersTanks[i]);
+			}
+		}
+		return listOfTanksWithoutCurrentTank;
+	}
+
 	render(): React.Node {
 		const cancelButton = (
 			<button className="cancelbtn" onClick={() => this.setState({popupOpen: false})}>Cancel</button>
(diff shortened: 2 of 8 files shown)


src/tanks/Tank.js
```diff
@@ -54,6 +53,7 @@ import {
 } from '../casus/userInteraction/CasusSpecialVariables.js';
 
 import type Battleground from '../battleground/Battleground.js';
+import ContainerBlock from '../casus/blocks/ContainerBlock.js';
```
SecondThread: This should be in the real imports section, not in the type imports section. (type imports get stripped out by the code we send to clients, so they don't actually cause JS cycles, that is why they are in a different section)

src/globalComponents/apiCalls/tankAPIIntegration.js
```diff
@@ -337,6 +339,41 @@ function getTanksById(tankIds: Array<string>, onLoad:(tanks: Array<Tank>) => voi
 	)
 }
 
+function updateCasusCode(tankId: string, casusCode: ContainerBlock): void {
```
SecondThread: This is duplicate code from `/casus/saveCasus.js`. You should probably make saveCasus pass in an optional tankId and make it only load from a cookie if the passed in id is null. Then we only need to add ~2 lines instead of ~50

src/armory/CopyCasusCodePopup.js
```diff
@@ -0,0 +1,95 @@
...
+			tankBeingCopiedFrom : null,
+		}
+	}
+
+	setTankBeingCopiedFrom(tankBeingUsedId: string): void {
+		let indexOfTankBeingUsed = -1;
+		for(let i = 0; i < this.props.usersTanks.length; i++) {
+			if(this.props.usersTanks[i]._id === tankBeingUsedId) {
+				indexOfTankBeingUsed = i;
+				i = this.props.usersTanks.length;
```
SecondThread: umm, what? Is this just a sketchy `break;`?

src/armory/CopyCasusCodePopup.js
```diff
@@ -0,0 +1,95 @@
...
+	constructor(props: Props) {
+		super(props);
+
+		this.state = {
+			popupOpen: false,
+			tankBeingCopiedFrom : null,
+		}
+	}
+
+	setTankBeingCopiedFrom(tankBeingUsedId: string): void {
```
SecondThread, line 32: I would double check to make sure you can't just make the value of these option tags just a tank object. I was under the impression that you could. That would make pretty much this entire method one line which would be very helpful.

src/armory/CopyCasusCodePopup.js
```diff
@@ -0,0 +1,95 @@
...
+			if(this.props.usersTanks[i]._id === tankBeingUsedId) {
+				indexOfTankBeingUsed = i;
+				i = this.props.usersTanks.length;
+			}
+		}
+		if(indexOfTankBeingUsed === -1) {
+			this.setState({tankBeingCopiedFrom: null});
+			return;
+		}
+		this.setState({tankBeingCopiedFrom: this.props.usersTanks[indexOfTankBeingUsed]});
```
SecondThread: If, for whatever reason, you can't just pass the tank object into this method, you can still clean it up quite a bit to something like:
`this.setState({tankBeingCopiedFrom: this.props.userTanks.find(tank => tank._id == tankBeingUsedId});`

SecondThread: Obviously if passing in the tanks works, you can do just:
`this.setState({tankBeingCopiedFrom: tankBeingUsed});`

src/armory/CopyCasusCodePopup.js
```diff
@@ -0,0 +1,95 @@
...
+							>
+								<option>Select A Tank To Copy From</option>
+								{this.props.usersTanks.map(tank => 
+									<option key = {tank._id} value = {tank._id}>{tank.tankName}</option>
+								)}
+							</select>
+							{copyButton}{cancelButton}
+						</div>
+					</Popup>
+					<ToastContainer />
```
SecondThread: We probably want this container just once on the outermost layer, right?

src/tanks/Tank.js
```diff
@@ -106,7 +106,7 @@ class Tank extends GameObject {
 
 	// Casus:
 	interpriterState: InterpriterState;
-	casusCode: CasusBlock;
+	casusCode: ContainerBlock;
```
SecondThread, line 113: Yeah, all of these Tank changes look substantially fine to me. As long as they pass flow that means we have been doing things consistently and it shouldn't be a problem for us to make decisions based on this being a ContainerBlock.

SecondThread (changes requested):
Mostly good. The two substantial changes are:
- we already have an api call for setting the casus code and we probably don't want a second one. (plus the existing one is really easy to modify to do what you need this to, so you should use that instead)
- setTankBeingCopiedFrom is unnecessarily long and complicated

src/armory/CopyCasusCodePopup.js
JorgeVidal29: If you do the value becomes [Object object] and not the actual object

SecondThread (changes requested):
The code looks fine. From a user perspective it needs work:
- The button is unreasonably large. It spans 3 lines compared to other 3 buttons next to it, which are all one line:
image
- The button is in the tank-options section, but it is a casus option, so it would make more sense to be by the casus button.
- The window doesn't close when you click copy.
- There is no confirmation asking the user if they are sure (this will overwrite all casus code in the previous tank). You could just delete a very large amount of work here if you select the wrong tank. You could implement this by needing to check a box saying "I understand this will overwrite all code in [whatever tank]" before the copy button is enabled

SecondThread (changes requested):
- The tank that you have selected in shown in the list which is confusing
- Just disable the "copy" button when the checkbox isn't clicked, don't show the user that they can click it and then give them an error when they do
- The Copy Code button should be light blue. Otherwise we have two primary buttons on one page.
- It is more clear if you use the word "overwriting" instead of "rewriting". You aren't writing the code again, you are deleting it and replacing it with something else
- Even if you were going to use the word "rewriting", you should spell it "rewriting" instead of "rewritting"

src/armory/Armory.js
```diff
@@ -324,10 +325,18 @@ class Armory extends React.Component<Props, State> {
...
 					<br/><br/>
+					<label>Edit Code</label>
+					<br/>
 					<Link to={verifyLink("/Casus")}>
 						<button className="primarybtn">Casus</button>
 					</Link>
-					<label>&emsp;Edit Code</label>
+					<br/><br/>
+					{selectedTank==null?<div></div>:
+							<CopyCasusCodePopup
```
SecondThread: fix indentation here, one too many tabs

src/armory/CopyCasusCodePopup.js
```diff
@@ -0,0 +1,120 @@
...
+			}
+		}
+		return listOfTanksWithoutCurrentTank;
+	}
+
+	render(): React.Node {
+		const cancelButton = (
+			<button className="cancelbtn" onClick={() => this.setState({popupOpen: false})}>Cancel</button>
+		);
+		const copyButton = this.state.confirmed===false?(<button className="smallbtn" disabled>Copy</button>):(<button className="smallbtn" onClick={() => this.copyCasusCode()}>Copy</button>);
```
SecondThread: also make sure that they have selected a tank before enabling

src/casus/saveCasus.js
```diff
@@ -5,9 +5,12 @@ import getTankForCasus from '../globalComponents/getTankForCasus.js';
 import ContainerBlock from '../casus/blocks/ContainerBlock.js';
 import { toast } from 'react-toastify';
 
-function saveCasus(casusCode: ContainerBlock): void {
+function saveCasus(casusCode: ContainerBlock, tankID: ?string, onLoad:() => void): void {
 	
-	const tankToEditID: string = getTankForCasus();
+	let tankToEditID: ?string = tankID;
```
SecondThread: You can clean this up to be: `const tankToEditID = tankID ?? getTankForCasus();`

SecondThread (approved):
Please fix changes before merging

PR #516, opened 2020-04-16
invalidate token if logged in twice

### LINKED ISSUE #477: Users can be logged in on two different websites at once.
**Describe the bug**
Isn't the verify route supposed to prevent this?

**To Reproduce**
1. Login.
2. Go to another computer.
3. Login.

**Expected behavior**
Should log the other person out because their JWT is no longer valid, correct?

**Screenshots**
N/A

**Additional context**
If this is good behavior, then close this issue. Although, I believe we didn't want this.

Airor4:
**Description**
Invalidates tokens from older sessions to enforce single user logins.

**Resolves These Issues**
Resolves #477 

**Type of change**

- [X] New feature (adds functionality)

1) Login user
2) make an authenticated api call and see if it is successful (it should be)
3) login in a new window and redo step 2
4) go back to the first login window and try to make an authenticated api call and it should no longer work.

**Final Checklist:**

- [X] My changes pass Flow, build without errors and (if applicable) pass Jest tests.
- [X] This change requires a update in the design document (if applicable)


SecondThread (approved):
lgtm

src/backend/controllers/userController.js
```diff
@@ -232,7 +232,8 @@ exports.login = async (req: Request, res: Response) => {
 		const payload = {
 			user: {
 				// Primary key id generated by MongoDB
-				id: user.id
+				id: user.id,
+				issued: Date.now()
```
Derplime, line 236: The issue here is that there is the small possibility that the `issued` property here and the `lastLogin` in the user might be _ever so slightly_ off, since they call separate instances of `Date.now()`

So before the logic that checks if you've logged in before to give a tank to the user, do a singular `const now = Date.now()` and then set everything that needs that `Date.now()` to `now`

Derplime (changes requested):
Just a quick nit on this. Other than that it's fine.

Derplime (approved):

Derplime:
For clarification, the nit is more of a suggestion to keep the issue date of the jwt in line with the lastLogin field.

element2112 (approved):

JohnJCordero (approved):

element2112 (approved):

PR #519, opened 2020-04-17
No more API calls when switching listings

### LINKED ISSUE #474: Going from a component view to tank view in Marketplace leads to a white screen.
**Describe the bug**
Read title.

**To Reproduce**
1. Go to Marketplace.
2. Click any of the component listings.
3. Click the tank listings.
4. View white screen.

**Expected behavior**
Should load the tanks correctly.

**Screenshots**
image

**Additional context**
N/A

### LINKED ISSUE #507: Marketplace crashes when a user sells their last component
**Describe the bug**
When selling the last component a user owns, the website will whitescreen.

**To Reproduce**
1. Sell all your components.

**Expected behavior**
Shouldn't crash.

**Screenshots**
image

**Additional context**
Might need to make a special case for if the components go to empty.

blaiadam:
**Description**
In Marketplace, we were making an API call and filtering all of the sales every time a user clicked a new listings view type. Now, it will get all of the market sales and split them up on the front end. Then, it will swap them out when a new listings view type is clicked. This PR should drastically speed up marketplace, as well as prevent crashes from swapping between types.

The Casus editor CSS change does not alter anything. The master listings header has been updated as well. Also, I fixed the crash on selling last component.

**Resolves These Issues**
Resolves #474 
Resolves #507 

**Type of change**

1. Make sure marketplace has tank sales and component sales.
2. Rapidly switch between them to make sure no crashes occur.
3. Purchase items to ensure that purchasing is still functioning as intended.

**Final Checklist:**


src/marketplace/ListingsView.js
```diff
@@ -11,19 +11,25 @@ import { allComponents } from '../globalComponents/typesAndClasses/TankComponent
...
+	scannersForSale: Array<SaleObject>,
+	scannerAddonsForSale: Array<SaleObject>,
+	jammersForSale: Array<SaleObject>,
+	treadsForSale: Array<SaleObject>,
 	itemsForSale: Array<SaleObject>,
-	// This is used to hold all of the tanks for sale. Will be null when the sellerType is not tank
-	tanksForSale: ?Array<Tank>
+	tanksForSale: Array<SaleObject>,
+	tanks: Array<Tank>,
+	cards: Array<React.Element<'div'>>
```
SecondThread: I haven't read everything yet, but this is almost always an indication that we are doing something wrong. We don't want to be storing how we are rendering something, we want to be storing the data that we are rendering, and then we can compute how to render it in the render() method

src/marketplace/ListingsView.js
```diff
@@ -32,117 +38,171 @@ class ListingsView extends React.Component<Props, State> {
...
 	}
 
-	// Once mounted, get the user's ID.
-	componentDidMount() {
-		this.getUserId();
+	// Once mounted, get the user's ID and set the sales.
+	componentDidMount(): void {
+		getUserAPICall(user => {
+			this.setState({userId: user.userId});
+			this.getSales();
```
SecondThread: getSales looks at this.state.userId, so shouldn't this be:
`setState({userId: user.userId}, this.getSales);`?

src/marketplace/ListingsView.js
```diff
@@ -32,117 +38,171 @@ class ListingsView extends React.Component<Props, State> {
...
-	// Else it directs it to getMarketSalesForComponents()
-	directSaleToProperFunction(): void {
-		if(this.props.sellerType === 'tanks') {
-			this.getMarketSalesForTanks();
-		}
-		else {
-			this.getMarketSalesForComponents();
+		if (prevProps !== this.props) {
+			switch (this.props.sellerType) {
+				case 'chassis':
```
SecondThread: Creating these cards should be done in the render method and they shouldn't be stored inside the component's state

src/marketplace/ListingsView.js
```diff
@@ -32,117 +38,171 @@ class ListingsView extends React.Component<Props, State> {
...
+			});
 
-	//This function uses the users id and gets the tanks that are active in the marketplace
-	getMarketSalesForTanks(): void  {
-		getMarketTanks(this.state.userId, sales => {
-				this.setState({itemsForSale: sales.filter(sale => !(allComponents.includes(sale.name)))}); 
-				this.getTanksToShow();
+			// After getting the sales, reset the cards according to the sellerType.
+			switch (this.props.sellerType) {
+				case 'chassis':
```
SecondThread: ditto earlier, and also this is a lot of unnecessarily duplicated code

src/marketplace/ListingsView.js
```diff
@@ -32,117 +38,171 @@ class ListingsView extends React.Component<Props, State> {
...
 
-	//This function uses tanks id from the state , creates those tanks , and adds them to the array of tanks
-	getTanksToShow() { 
-		const arrayOfTankIds = [];
-		for(let i = 0; i < this.state.itemsForSale.length; i++) {
-			if(this.state.itemsForSale[i].tankId == null) {
+	// Converts SaleObject to Tank.
+	convertSalesToTanks(saleTanks: Array<SaleObject>): void {
+		// Do not convert if there are no tanks to convert.
+		if (saleTanks.length === 0) {
```
SecondThread: why do we need this special case? If our code doesn't work when there are no tanks to convert it seems like we have messy logic somewhere...

src/marketplace/ListingsView.js
```diff
@@ -32,117 +38,171 @@ class ListingsView extends React.Component<Props, State> {
...
 		});
 	}
 
-	// This creates a card for every sale
-	createCards = () => {
-		const cards = []
-		//Check for no sales
-		if(this.state.itemsForSale.length === 0) {
-			return(<h5>No Sales Availiable At This Time</h5>);
+	createCards = (itemsForSale: Array<SaleObject>, saleType: MarketplaceViewType): Array<React.Element<'div'>> => {
```
SecondThread: This method should only ever be called from render()

src/marketplace/Marketplace.js
```diff
@@ -23,57 +23,59 @@ class Marketplace extends React.Component<Props, State> {
 	constructor() {
 		super();
 		verifyLogin();
-		this.state={
-			marketplaceViewClicked: 'NONE'
+		this.state = {
+			marketplaceViewClicked: null
```
SecondThread, line 27: thanks

SecondThread (changes requested):
The big change I am requesting is that this is storing the _way_ that things are being rendered in the state, instead of storing _what_ is being rendered. This leads to duplicated code, states that could be illegal (if how it is being rendered doesn't match what is being rendered), and confusion when other people try to understand what is going on.

src/marketplace/ListingsView.js
blaiadam: Good catch.

src/marketplace/ListingsView.js
blaiadam: Once I am back, I will change it to map the cards out inside the render method instead. It might take me a bit to change stuff around to make it work, but it can't be too hard.

JorgeVidal29 (approved):
Passed flow, was able to see listings, and was able to buy and sell items. Looks good to me

SecondThread (approved):
Looks good, thanks for changing

PR #524, opened 2020-04-19
Marketplace headings cleanup

### LINKED ISSUE #515: Fix headings on marketplace and sort them
Sort by: if one is from factory and one isn't put the one from a player on top.
If they are both from factory or both from player, sort by price.

**Additional context**

SecondThread:
**Description**
Cleaned up the headings of the marketplace tabs. Most of it was already done, so there was very little I had to actually do here. Mostly it just involved: 
- adjusting the id of the master account
- sorting items in a reasonable order (sorted by price/unit, with items that are sold by other players on top)

Screenshot of what it looks like:
image

The changes were also already implemented for tanks, so I didn't have to do anything there.

**Resolves These Issues**
Resolves #515 

**Type of change**

**Final Checklist:**

### DIFF
--- a/src/globalComponents/getMasterAccountId.js
+++ b/src/globalComponents/getMasterAccountId.js
@@ -1,0 +1,9 @@
+//@flow strict
+
+const MASTER_ACCOUNT_ID = '5e7ff4a9c6c3700959497009';
+
+function getMasterAccountId(): string {
+	return MASTER_ACCOUNT_ID;
+}
+
+export default getMasterAccountId;
--- a/src/globalComponents/getMasterAccountId.js
+++ b/src/globalComponents/getMasterAccountId.js
@@ -1,6 +1,6 @@
 //@flow strict
 
-const MASTER_ACCOUNT_ID = '5e7ff4a9c6c3700959497009';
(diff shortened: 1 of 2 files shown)


src/globalComponents/getMasterAccountId.js
```diff
@@ -1,6 +1,6 @@
 //@flow strict
 
-const MASTER_ACCOUNT_ID = '5e7ff4a9c6c3700959497009';
+const MASTER_ACCOUNT_ID = '5e93b1d0d1125f22ecd469b7';
```
Derplime, line 3: Is it possible for you to grab it from the `.env`? The master account ID is in that file. If you can't then that's fine. I'm just uncomfortable for having the master ObjectId hardcoded here. Also its convenient so we or future groups just need to update the master ID in one place.

Just make sure you update the `.env` file in the Drive if you make this change.

Derplime (approved):
Have a suggestion here, otherwise it's good.

src/globalComponents/getMasterAccountId.js
SecondThread: I'm pretty sure we can't/wouldn't want to. If we include the env file in what we send to the frontend, then we are sending them all of our passwords/keys in plaintext too, which is bad

PR #539, opened 2020-12-29
Docker

element2112:
Split application into isolated folders that represent the backend and frontend. Added support for docker


JohnJCordero (approved):
looks good

PR #540, opened 2021-01-14
Casus editor layout

element2112:
**Description**
Changed casus layout/added new font/added bg/fixed rendering bug for tank display


JohnJCordero (approved):
Reviewed all files, all edits look fine, reminder that Navbar is now MainNavBar in case of merge conflicts

PR #543, opened 2021-01-29
Tutorial joyride

element2112:
**Description**
Adding joyride to main page/armory/training arena

Content fields with descriptions will be filled in by john once api call is added


JohnJCordero (approved):
Everything looks good, all added content was for JoyRide ; blocks were added and blockShelf css was added

PR #568, opened 2021-03-30
changed server port to expose

element2112:
changed server port to 5000 in dockerfile


JohnJCordero (approved):
changed port for docker

